PR #33, opened 2017-04-22
Test token

janczer:
Add test to token.go


st0012:
@janczer Thanks for the PR, you're the first contributor of Rooby 🎉🎉🎉

Issue #37, opened 2017-04-23
Add line numbers to AST node


Heasummn:
I'd like to help with the development of rooby, and this feature seems pretty simple to implement. I was wondering if I could take a go at it.

PR #38, opened 2017-04-23
Added Line() function to all AST nodes

Heasummn:
Closes issue #37


st0012:
@Heasummn Can you please add some tests for this change? thanks 😄

st0012:
@Heasummn Hi, If you want your branch's base up to date, you should update master branch to the latest commit, and then checkout to this branch and `rebase master`. Merging master branch is not a good practice unless you have good reason for doing it~

hzlmn:
```golang
func (_ *Program) Line() int {
 	// The program always starts at line 1
	return 1
 }
```
I think it is wrong as lexer support skipping whitespaces program can start not exactly on 1 line.
@st0012  Isn't it?

st0012:
@hzlmn Agree, we need to consider that there would be some files contain some empty lines before the first line of code. @Heasummn Can you help change that and add some test in `parser` package? thx

st0012:
@Heasummn Hi, are you still working on this?

PR #44, opened 2017-04-25
Add a gitattributes file to highlight .ro files as ruby on GitHub

adelcambre:
GitHub uses linguist for language detection, which supports `.gitattributes` files for hinting. https://github.com/github/linguist#using-gitattributes

This should make `*.ro` files show up syntax highlighted with the Ruby parser on GitHub.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ro linguist-language=ruby
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-*.ro linguist-language=ruby
+*.ro linguist-language=Ruby


st0012:
@adelcambre Thanks for this PR, I never thought I could do something like this 😂. But I visited your fork and it seems like `*.ro` files still have no syntax highlight?

adelcambre:
@st0012 On further investigation, it looks like using gitattributes to set syntax languages like this isn't supported on github.com yet. (I noticed it wasn't working on my fork and thought it was just cause it was on a branch or something.) I'll go ahead and close this.

adelcambre:
Actually, it looks like it does work! As of very recently! I think maybe it doesn't work with forks or something but I just pushed a test to a test repo and it's working there: https://github.com/adelcambre/test/blob/master/rooby/test.ro

adelcambre:
This is where it says it's working: https://github.com/github/linguist/issues/1792#issuecomment-286379822

st0012:
@adelcambre Ok, thank you! I will merge and give it a shot 😁.

st0012:
@adelcambre Just tested on master and it works, Good Job! However, it won't automatically apply to existing files, only after they are changed.

adelcambre:
@st0012 I think it reprocesses the existing files on push, but it might take a few minutes. At least that's what it says in https://github.com/github/linguist/issues/1792#issuecomment-286379822

PR #46, opened 2017-04-25
Fix golint warnings

SumLare:


st0012:
@SumLare Hi, thanks for you contribution! And I think it would be better if we change all the constant names to make them consistent. For example `STRING` should be `String`, `LPAREN` to `LParen`. What do you think?

SumLare:
@st0012 Sounds relevant. I thought about it, will update it soon.

st0012:
@SumLare LGTM, thanks!

PR #47, opened 2017-04-27
Add new method to Array and Integer

janczer:
Add method 'map', 'select' to Array. Add method 'even', 'odd' to Integer

### DIFF
--- a/vm/array_test.go
+++ b/vm/array_test.go
@@ -165,19 +165,19 @@ func TestMapMethod(t *testing.T) {
 		a.map do |i|
 			i + 3
 		end
-		`, InitializeArray([]Object{initilaizeInteger(4), initilaizeInteger(5), initilaizeInteger(10)})},
+		`, initializeArray([]Object{initilaizeInteger(4), initilaizeInteger(5), initilaizeInteger(10)})},
 		{`
 		a = [true, false, true, false, true ]
 		a.map do |i|
 			!i
 		end
-		`, InitializeArray([]Object{FALSE, TRUE, FALSE, TRUE, FALSE})},
+		`, initializeArray([]Object{FALSE, TRUE, FALSE, TRUE, FALSE})},
 		{`
 		a = ["1", "sss", "qwe"]
 		a.map do |i|
 			i + "1"
 		end
-		`, InitializeArray([]Object{initializeString("11"), initializeString("sss1"), initializeString("qwe1")})},
+		`, initializeArray([]Object{initializeString("11"), initializeString("sss1"), initializeString("qwe1")})},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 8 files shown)


vm/vm_test.go
```diff
@@ -451,6 +451,37 @@ func testBooleanObject(t *testing.T, obj Object, expected bool) bool {
...
 
+func testArrayObject(t *testing.T, obj Object, expected *ArrayObject) bool {
+	result, ok := obj.(*ArrayObject)
+	if !ok {
+		t.Errorf("object is not Array. got=%T (%+v)", obj, obj)
+		return false
+	}
+
+	for i := 0; i < len(result.Elements); i++ {
+		into, ok := expected.Elements[i].(*IntegerObject)
```
st0012: I think `intObj` would be better than `into`

vm/boolean.go
```diff
@@ -39,6 +39,10 @@ func (b *BooleanObject) ReturnClass() Class {
 	return b.Class
 }
 
+func initializeBoolean(value bool) *BooleanObject {
```
st0012: We should not have a method to initialize boolean. I already use variable `TRUE` and `FALSE` to represent true and false. Since they won't be modified, we can share them between the program without creating new one.

vm/array_test.go
```diff
@@ -155,6 +156,110 @@ func TestEachMethod(t *testing.T) {
...
+	tests := []struct {
+		input    string
+		expected *ArrayObject
+	}{
+		{`
+		a = [1, 2, 3, 4, 5]
+		a.select do |i|
+			i > 3
+		end
+		`, InitializeArray(elementsInt)},
```
st0012: I think you should generate array with clear arguments. For example in this case it should be
```go
InitializeArray(initilaizeInteger(4), initilaizeInteger(5))
```
I know this seems very verbose but I can now know the expected result just by watching this case without scrolling. And this can be further refactored in the future. Same rule applies to other tests.

st0012:
@janczer Hey thanks for the PR, good job 😄. Just left a few comments about what can be improved.

janczer:
All changes after code review I add to one commit.

vm/integer.go
```diff
@@ -282,6 +282,43 @@ var builtinIntegerMethods = []*BuiltInMethod{
 		},
 		Name: "to_s",
 	},
+	{
+		Fn: func(receiver Object) builtinMethodBody {
+			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
+				if len(args) > 0 {
+					return &Error{Message: "Too many arguments for Integer#--"}
```
st0012: You forgot to change the error message~

st0012:
@janczer BTW can you rebase it before merge?

vm/integer.go
janczer: I coped it from `to_s` method. What message should be?

st0012: `Integer#even`
This is Ruby's convention, we declare method name using `ClassName#MethodName`. So the below should be `Integer#odd`

st0012: And that also means the `to_s`'s message is wrong.

st0012:
@janczer I unexported `InitializeArray` method so you need to downcase the first character `I` to `i`

janczer:
I fixed it, and fix in `method.go` `vm.stack.Top` to `vm.stack.top`.

PR #52, opened 2017-04-28
Add new method: <=, >= and <=>

janczer:


lexer/lexer.go
```diff
@@ -61,11 +61,35 @@ func (l *Lexer) NextToken() token.Token {
...
+		}
 	case '<':
-		tok = newToken(token.LT, l.ch, l.line)
+		if l.peekChar() == '=' {
+			currentByte := l.ch
+			l.readChar()
+			if l.peekChar() == '>' {
+				currentByteNext := l.ch
+				l.readChar()
+				tok = token.Token{Type: token.COMP, Literal: string(currentByte) + string(currentByteNext) + string(l.ch), Line: l.line}
```
st0012: If you know this is a compare operator, why not just write `<=>` as literal?

st0012: Same rule applies to other new operators

janczer:
Add test for `**`. But don't know if it's enough?

vm/integer_test.go
```diff
@@ -27,6 +27,19 @@ func TestEvalInteger(t *testing.T) {
...
+		{`4 > 6`, false},
+		{`4 >= 4`, true},
+		{`2 >= 5`, false},
+		{`-5 < -4`, true},
+		{`100 < 81`, false},
+		{`10 <= 10`, true},
+		{`10 <= 0`, false},
+		{`10 <=> 0`, 1},
+		{`1 <=> 2`, -1},
+		{`4 <=> 4`, 0},
```
st0012: Can you add more complicated arithmetic tests in here? Which should be a combination of different operators.

vm/integer.go
```diff
@@ -186,6 +238,62 @@ var builtinIntegerMethods = []*BuiltInMethod{
...
+				leftValue := receiver.(*IntegerObject).Value
+				right, ok := args[0].(*IntegerObject)
+
+				if !ok {
+					return wrongTypeError(integerClass)
+				}
+
+				rightValue := right.Value
+
+				if leftValue < rightValue {
```
st0012: In Go we prefer
```go
if leftValue < rightValue {
        return initilaizeInteger(-1)
}

if leftValue > rightValue {
        return initilaizeInteger(1)
}

return initilaizeInteger(0)

Comment (st0012) [inline vm/integer.go]: If you run `golint` on this file, it should raise a warning about it.

File: vm/integer.go [hunk 4]
```diff
@@ -159,6 +184,33 @@ var builtinIntegerMethods = []*BuiltInMethod{
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 
+				err := checkArgumentLen(args, integerClass, ">=")
```
st0012: As we discuss earlier, you can remove those argument length checks.

PR #58, opened 2017-05-02
Add suport operators && and ||

janczer:
Relate to #55

### DIFF
--- a/vm/boolean_test.go
+++ b/vm/boolean_test.go
@@ -49,6 +49,10 @@ func TestEvalInfixBooleanExpression(t *testing.T) {
 		{"false || true", true},
 		{"true || false", true},
 		{"false || false", false},
+		{"100 > 10 && true == false", false},
+		{"true && true == true", true},
+		{"(false || true) && (\"string\" == \"string\")", true},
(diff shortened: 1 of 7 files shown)


vm/boolean_test.go
```diff
@@ -41,6 +41,14 @@ func TestEvalInfixBooleanExpression(t *testing.T) {
...
 		{"(1 > 2) == true", false},
 		{"(1 > 2) == false", true},
+		{"true && true", true},
+		{"false && true", false},
+		{"true && false", false},
+		{"false && false", false},
+		{"true || true", true},
+		{"false || true", true},
+		{"true || false", true},
+		{"false || false", false},
```
st0012: @janczer Can you add more complex condition to test new precedence setting? Like:

```ruby
100 > 10 && true == false # false
```

st0012:
LGTM, just need more tests for new precedence 😄

janczer:
How I can test new precedence?

vm/boolean_test.go
st0012: @janczer Here's the example for testing precedence.

janczer: @st0012 Yes, I can. Thank you

PR #67, opened 2017-05-03
Add api_reference example

adlerhsieh:
I believe API documents are good for developers who want to contribute to this project. 

This is an example of documenting API from core libraries. Since this project is not complicated for now, we can take a MVP approach to generate simple markdown documents like this. When the API becomes so complicated that it requires new format, then it can be moved to another place, maybe self-hosting or using GitHub Pages. We can discuss which format is better. 

A better approach than copying methods from the source code is to write a parser. That parser will parse the source code and generate markdown or HTML files automatically, like Rails and its API pages. That can be done after this.

You can find this example displayed on [my forked repo](https://github.com/adlerhsieh/rooby/tree/api_reference/api_reference).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ I want to build a language that focuses on developing microservices. Which shoul
 A lot people have questions about `rooby` since it's a new language and you may get confused by the way I describe it (sorry for that 😢). Here's a list of [frequently asked questions](https://github.com/rooby-lang/rooby/wiki/Frequently-asked-questions).
 
 ## Supported features
+
+#### [API Reference](api_reference/)
+
 - **Can be compiled into bytecode (with `.robc` extension)**
 - **Can evaluate bytecode directly**
 - Everything is object
--- a/api_reference/README.md
+++ b/api_reference/README.md
@@ -1,0 +1,5 @@
+# API Reference
+
+(WIP)
+
+- [Integer](integer.md)
--- a/api_reference/integer.md
+++ b/api_reference/integer.md
@@ -1,0 +1,83 @@
+# Integer
+
+[Source code](https://github.com/rooby-lang/rooby/blob/master/vm/integer.go)
+
+### + (Integer)
+
+Returns the sum of self and another `Integer`.
+
+### - (Integer)
+
+Returns the subtraction of another `Integer` from self.
+
+### * (Integer)
+
+Returns self multiplying another `Integer`.
+
+### ** (Integer)
+
+Returns self squaring another `Integer`.
+
+### / (Integer)
+
+Returns self divided by another `Integer`.
+
+### > (Integer)
+
+Returns if self is larger than another `Integer`.
+
+### >= (Integer)
+
+Returns if self is larger than or equals to another `Integer`.
+
+### < (Integer)
+
+Returns if self is smaller than another `Integer`.
+
+### <= (Integer)
+
+Returns if self is smaller than or equals to another `Integer`.
+
+### <=> (Integer)
+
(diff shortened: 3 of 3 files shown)


st0012:
@adlerhsieh I think it's a good idea to have an API document from now. However, we are currently busy at developing features and have no time to write or maintain this document 😅

adlerhsieh:
I agree that leaving documents incomplete is a bad idea, so what I'm trying to do is give an example on how this document can look like. If you are feeling ok with this format I can help work on a parser I mentioned, so that you don't have to maintain a document. Instead, every time you add something, run a command like 'make doc' to automatically generate docs for current libraries.

st0012:
I think the format is good enough for me, except the argument's type in `+`, `-`...etc. seems a little unnecessary and verbose? @janczer What do you think?
But I think if you can make a doc generator it will help us alot. And before you begin can you tell us what format should we document those APIs? BTW I can't guarantee the way we create built in methods won't be change (this might effect your parser) since we're at early stage now.

st0012:
And if you're ready to start the document parser, I will open a new repo for it and add you as a collaborator so you can work on it directly.

janczer:
I think will be better add some examples of methods:
```
3.next #=> 4
```

adlerhsieh:
Yes, there's a risk that things may change in the future. I'll do some research on your current pattern, looking for possibilities to create a parser without too much hard coding and tweaks. Also it's worth having a discussion on how these methods are documented in the code. I'll come back when it's ready.

adlerhsieh:
I think this part is ready. The documenting approach is as follow:

## Class

Each file under `/vm` directory will be parsed. A class will be documented if:

- There is a `Xxx` or `XxxObject` type declaration in the file name. For example, `BooleanObject` represents `Boolean` class in `boolean.go`, and `Null` represents `Null` class in `null.go`.
- `*_test.go` files will be excluded.
- The documentation of the class is at right above the declaration, like:

```go
// BooleanObject represents boolean object in Rooby
type BooleanObject struct {
	Class *RBool // field comments
	Value bool
}
```

## Instance Method

If a class is found and documented by the parser, it will look for instance methods by the rules:

- The declaration of `builtInXxxMethods` must exists. For example, `builtinIntegerMethods` declaration will be parsed if there is a `Integer` class in `integer.go` file.
- The documentation of a method is right above the `Fn` key. It is not above the bracket `{`. For example:

```go
	{
		// Returns the sum of self with another value.
			// ...
		},
		Name: "+",
	},
}
```

## Next Step

If you are fine with this approach, you can create a repository for the parser, maybe called "doc-parser", "doc-generator", or simply "doc". I'll add code there and generate some `.md` files. 

Maybe after that we can discuss if some HTML files should be generated and integrated into your [existing index page](https://rooby-lang.github.io/rooby/).

st0012:
@adlerhsieh This looks good to me. But I think we should add example comment for each method like:

```go
	{
		// Returns the sum of self with another value.
                // Example: 1 + 1
			// ...
		},
		Name: "+",
	},
}
```
so we can have example for our methods.

Another thing is that I know `rooby`'s current naming rule is not very consistent, if you think change any struct's name can help building this parser please let me know (like change `Null` to `NullObject`). With IDE this won't be a hard task.

adlerhsieh:
That won't be a problem if we want to add multi-line comments.

For naming the parser is able to handle both cases easily, but I agree for consistency it is better to change some exceptions like Null and Object. However that could be done in a separate issue.

st0012:
@adlerhsieh Sure. BTW can you check gitter, I have another issue that needs help 😅

adlerhsieh:
Closing this one and move to #85

PR #69, opened 2017-05-04
add String#to_i & Integer#to_i

adlerhsieh:
`String#to_i` implementation resembles Ruby.

### DIFF
--- a/vm/integer.go
+++ b/vm/integer.go
@@ -320,10 +320,7 @@ var builtinIntegerMethods = []*BuiltInMethod{
 	{
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
-
-				int := receiver.(*IntegerObject)
-
-				return initilaizeInteger(int.Value)
+				return receiver
 			}
 		},
 		Name: "to_i",
--- a/vm/integer.go
+++ b/vm/integer.go
@@ -321,6 +321,17 @@ var builtinIntegerMethods = []*BuiltInMethod{
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 
+				int := receiver.(*IntegerObject)
+
+				return initilaizeInteger(int.Value)
+			}
+		},
+		Name: "to_i",
+	},
+	{
+		Fn: func(receiver Object) builtinMethodBody {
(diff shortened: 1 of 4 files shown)


vm/integer.go
```diff
@@ -321,6 +321,17 @@ var builtinIntegerMethods = []*BuiltInMethod{
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 
+				int := receiver.(*IntegerObject)
+
+				return initilaizeInteger(int.Value)
```
st0012: Why initialize a new integer instead of returning the original one?

vm/string.go
```diff
@@ -297,6 +299,36 @@ var builtinStringMethods = []*BuiltInMethod{
...
+			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
+
+				str := receiver.(*StringObject).Value
+				parsedStr, err := strconv.ParseInt(str, 10, 0)
+
+				if err == nil {
+					return initilaizeInteger(int(parsedStr))
+				}
+
+				var digits string
```
st0012: Wow so this is how Ruby implement `String#to_i`?

adlerhsieh: Sorry, not implementation, I mean the outcome. Go's `strconv.ParseInt` returns 0 if parsing strings that contain any non-digit character like `"123string"`. 

However, Ruby's `String#to_i` method goes through the string and returns all digits before the first a non-digit character. For example, `"123string".to_i` would return `"123"`, and `"str123"` would return `0`.

vm/integer.go
adlerhsieh: Good point.

vm/integer.go
```diff
@@ -321,6 +321,17 @@ var builtinIntegerMethods = []*BuiltInMethod{
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 
+				int := receiver.(*IntegerObject)
```
st0012: I think just `return receiver` should work

vm/string.go
st0012: Ok, thanks for the explain, I didn't know this before XD

adlerhsieh:
Updated

st0012:
@adlerhsieh Thanks ^_^

PR #70, opened 2017-05-04
Support module

st0012:


st0012:
@janczer Can you help me review the code?

vm/statement_test.go
```diff
@@ -116,6 +116,56 @@ func TestDefStatement(t *testing.T) {
...
+		end
+
+		class Foo
+		  include(Bar)
+		end
+
+		Foo.new.bar
+		`,
+			10},
+		{`module Foo
```
st0012: @adlerhsieh BTW, another thing you can contribute is to design some more complex tests for `rooby` 😀

vm/class.go
```diff
@@ -60,12 +61,16 @@ type BaseClass struct {
 	Methods *environment
 	// ClassMethods contains this class's methods
 	ClassMethods *environment
-	// SuperClass points to the class it inherits
-	SuperClass *RClass
+	// pseudoSuperClass points to the class it inherits
+	pseudoSuperClass *RClass
+	// This is the class where we should looking for a method.
+	// It can be normal class, singleton class or a module.
+	superClass *RClass
```
janczer: Why we have two instance `pseudoSuperClass` and `superClass`? In `Integer`, `String` etc this variables have the same object.

st0012: In Ruby, module is almost like class internally and module including represents as inheritance, so we need to know the actual superclass (might be a module), and the superclass for checking class inheritance. For example:
```ruby
module Foo
end

class Baz
end

class Bar < Baz
  include(Foo)
end
```

In the internal, the inheritance looks like `Bar < Foo < Baz` (record is `superClass` attribute)
But we need to know the class inheritance `Bar < Baz` so we also need `pseudoSuperClass`

PR #75, opened 2017-05-06
Improve performance

st0012:
Currently `rooby`'s performance is not very ideal. For example:
```ruby
i = 1000000
while i > 0; do
    i = i - 1
end
```
This simple loop iteration require about `100ms` for Ruby but almost 3 seconds for `rooby`(until this commit https://github.com/rooby-lang/rooby/commit/2560a6aa953edc0837317d77046209bf1e4fba08).
So I and @janczer did some research using a [profile package](https://github.com/pkg/profile) and found that there are a few reason:

- We use terrible regex to delete newline character
- `Sprintf` can slow done the program if we put it in some frequently iterated section (not sure yet)

For now we reduced the iteration time to about 1 second, and will keep work on it.


vm/instruction.go
```diff
@@ -283,7 +283,7 @@ var builtInActions = map[operationType]*action{
 				constant := vm.constants[constantName]
 				inheritedClass, ok := constant.Target.(*RClass)
 				if !ok {
-					newError("Constant %s is not a class. got=%T", constantName, constant)
+					panic("Constant "+ constantName +" is not a class. got=" + string(constant.Target.objectType()))
```
janczer: Why are you using `panic`? Maybe it would be better to use `&Error{}`?

vm/integer.go
```diff
@@ -312,7 +311,7 @@ var builtinIntegerMethods = []*BuiltInMethod{
 
 				int := receiver.(*IntegerObject)
 
-				return initializeString(fmt.Sprint(int.Value))
+				return initializeString(string(int.Value))
```
janczer: You should use `strconv.Itoa`

vm/instruction.go
st0012: @janczer Because `newError` returns an error object, but we didn't return it here so it didn't work at all.  But I will change it to pushing error to the stack.

vm/integer.go
st0012: @janczer What will be the difference?

st0012:
@janczer Can you review again?

PR #85, opened 2017-05-13
API Documentation Parser: Integer Class Example

adlerhsieh:
This pull request is also or reviewing the API parser and generator. Currently I use Integer class as an example:

https://goby-lang.github.io/api.doc/integer.html

Please also check for the approach for documentation on [README](https://github.com/goby-lang/api.doc#documenting-code).


adlerhsieh:
I'm not quite sure what Codeclimate means. I use `gofmt` on the file and it does not change anything. Could you help explain? Thanks.

st0012:
@adlerhsieh This looks awesome! And since standard libraries will have similar file structure, I think this can apply to them too.

BTW I think you can also explain how it finds `goby`'s `vm` folder in readme. Because it's a little magical that we can run a command in api project's root and get the data we want from other project.

(I don't know what the issue is either so I approved that check.)

PR #89, opened 2017-05-15
Fix typo

hachi8833:
My first PR
I hope this works.


st0012:
@hachi8833 Thanks for the PR, but that's not typo, see [here](https://ruby-doc.org/core-2.2.0/Integer.html#method-i-pred).

hachi8833:
Sorry, I should've checked that.

PR #90, opened 2017-05-15
Refactor method definitions in vm/ for readability

hachi8833:
I hope this change might be good this time.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -91,6 +91,7 @@ func init() {
 
 var builtinArrayMethods = []*BuiltInMethod{
 	{
+		Name: "[]",
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 				if len(args) != 1 {
@@ -114,9 +115,9 @@ var builtinArrayMethods = []*BuiltInMethod{
 
 			}
 		},
-		Name: "[]",
 	},
 	{
+		Name: "[]=",
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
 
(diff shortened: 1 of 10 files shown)


st0012:
Moving method name up is a good idea, but I'm not sure if this would effect document generation. @adlerhsieh can you check this?

adlerhsieh:
Currently it will not effect the parser, although the style might need some tweak. If putting `Name` above `Fn`, then the comment for this method still needs to come right above `Fn`. For example:

```go
  {
    Name: "+",
    // Returns the sum of self and another Integer.
      // ...
    },
}
```

But we're still in an early stage so changing this style won't take too much time. Comments can come either above `Name` or  bracket `{`.

st0012:
@adlerhsieh I think you can decide the style, although I prefer comment upon the name.

hachi8833:
Updated my master branch with the style as adlerhsieh suggested.
(I was just wondering how to update the PR yesterday)

adlerhsieh:
To update a PR, just push to the branch again. In you case, push to `master` will do. 

However, some general guidelines to follow if you want to keep contributing:

- Checkout to a new branch when working on a new feature. When opening a PR, it means your things are stuck with that branch. Sometimes a PR could take a week of discussion before merge. If you have something new in mind, you won't be able to progress if you always work on `master`. In your case, modifying `Integer#prev` including 0688a04a5288e4f72b69910e608de232b0aa064f and c836f7466e89652bc5cacbb3c9d687edfb6e1160 are not related to this PR.
- Use `git commit --amend` and `git push --force` to overwrite commits. If you add a line and commit, and after discussion you add another line and commit. In my experience doing this 10 times in a PR is common, and it makes the commits very dirty. It's better to overwrite the same stuff.

hachi8833:
Sorry, I should've PR with a new feature branch, instead of merging to my local master branch. I'll be careful next time.

PR #91, opened 2017-05-16
Move method comments to the top of method name

adlerhsieh:
I've updated the parser so a method structure now looks like this:

```go
  {
    // Returns the sum of self and another Integer.
    //
    // ```
    // 1 + 1
    // ```
    Name: "+",
      // ...
    },
  },
}
```

Based on this change I updated the existing doc in `vm/integer.go`


st0012:
@adlerhsieh ok thanks!

PR #92, opened 2017-05-16
Allow boolean type to compare with other types

adlerhsieh:
It turns out that comparisons between types can be done through `interface{}` according to Go's [doc](https://golang.org/ref/spec#Comparison_operators).

> Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.

And that's what we're actually doing, so the implementation can be simplified.


st0012:
@adlerhsieh This is a huge improvement, awesome!

PR #95, opened 2017-05-16
Refactor function format in vm

hachi8833:
Updated the test and I'd try again.

I hope this does not disturb your team again.

Thank you,


vm/array.go
```diff
@@ -77,7 +77,7 @@ func initializeArray(elements []Object) *ArrayObject {
 	return &ArrayObject{Elements: elements, Class: arrayClass}
 }
 
-func init() {
+func initArray() {
```
st0012: The init method is Go's built in method that will be executed when the package is imported.

st0012:
Wow this is a huge work, thanks! But we currently still under rapid development, so large refactoring is too early, we should do this after the core functionality is stable. I'm closing it for now, sorry didn't make it clear earlier 😢

vm/array.go
hachi8833: OK, I understood that using `init()` is intentional 😺 
I just thought it looked inconsistent with other initialization functions.

hachi8833:
No problem, I understood your situation 😺 
I just thought it might be good to reduce the repetitive codes in early stage.
I'd do that in a good chance some day!

PR #104, opened 2017-05-20
Add Parser Options

adlerhsieh:
The following options are available when running `goby`:

- `-v` display version
- `-h` display help message

Also fix the error when no filepath is specified. The program now displays the help message and exit.

### DIFF
--- a/goby.go
+++ b/goby.go
@@ -35,6 +35,12 @@ func main() {
 	}
 
 	filepath := flag.Arg(0)
+
+	if filepath == "" {
+		flag.Usage()
+		os.Exit(0)
+	}
+
 	args := flag.Args()[1:]
 
 	dir, filename, fileExt := extractFileInfo(filepath)
--- a/goby.go
+++ b/goby.go
@@ -3,26 +3,35 @@ package main
 import (
 	"flag"
 	"fmt"
-	"github.com/pkg/profile"
 	"github.com/goby-lang/goby/bytecode"
 	"github.com/goby-lang/goby/parser"
 	"github.com/goby-lang/goby/vm"
+	"github.com/pkg/profile"
 	"io/ioutil"
+	"log"
 	"os"
 	"path"
-	"strings"
-	"log"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


goby.go
```diff
@@ -3,29 +3,59 @@ package main
...
+  -h   show this help message
+
+For more details, visit: https://github.com/goby-lang/goby
+`
+
 func main() {
 	compileOptionPtr := flag.Bool("c", false, "Compile to bytecode")
 	profileOptionPtr := flag.Bool("p", false, "Profile program execution")
+	versionOptionPtr := flag.Bool("v", false, "Show current Goby version")
+	helpOptionPtr := flag.Bool("h", false, "Show help message")
```
st0012: Actually you don't have to add this, the flag package already takes care this for us. You can try `goby -h` on master branch.

goby.go
```diff
@@ -3,29 +3,59 @@ package main
...
 	"path"
-	"strings"
-	"log"
 	"path/filepath"
+	"strings"
 )
 
+// Version stores current Goby version
+const Version string = "0.0.1"
+
```
st0012: This reminds me that we should also create a procedure for bumping versions.

goby.go
adlerhsieh: Thanks. I didn't know that.

goby.go
adlerhsieh: You mean something other than `-v` flag?

adlerhsieh:
Updated. It now uses `flag.Usage()` when no file is specified.

goby.go
st0012: No, I mean we need to document about updating this constant, create a new release, update brew formula...etc.

st0012: @adlerhsieh Just update added it into the first release's todo list.

st0012:
@adlerhsieh Ok thanks!

goby.go
adlerhsieh: Yeap, sure. We will need that for maintainers.

PR #105, opened 2017-05-20
Minor improvements on vm package

st0012:
Remove some useless code and refactor some methods.


adlerhsieh:
I'm not quite familiar with the whole design, but I agree that we have to implement error system at some point. Looks good to me.

PR #110, opened 2017-05-24
Update README for new contributors

hachi8833:
Just to indicate `make test` is required before PR.
(I fell the same pitfall)


st0012:
@hachi8833 Thanks, I just assume that everybody knows that, clearly I'm wrong 😅. This is a good reminder.

PR #118, opened 2017-05-26
HTTP library

st0012:
Add `Net::HTTP` class and `URI` module

### DIFF
--- a/vm/http.go
+++ b/vm/http.go
@@ -20,6 +20,7 @@ func initializeHTTPClass(vm *VM) {
 
 var builtinHTTPClassMethods = []*BuiltInMethod{
 	{
+		// Sends a GET request to the target and returns the HTTP response as a string.
 		Name: "get",
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(v *VM, args []Object, blockFrame *callFrame) Object {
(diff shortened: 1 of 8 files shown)


vm/http.go
```diff
@@ -0,0 +1,50 @@
...
+	for _, m := range builtinHTTPClassMethods {
+		http.ClassMethods.set(m.Name, m)
+	}
+
+	vm.constants["Net"] = &Pointer{Target: net}
+}
+
+var builtinHTTPClassMethods = []*BuiltInMethod{
+	{
+		Name: "get",
```
janczer: Maybe should add description to methods?

st0012: I'll add it after I finish them 😄

st0012:
@adlerhsieh I agree that it seems to be redundant. But since most languages (at least in Go and Ruby)  has this namespace, I think maybe we just haven't know the reason, or that just became a convention.

adlerhsieh:
Yes, I agree.

st0012:
I found that `Net::HTTP` library in ruby is far more complicated than I thought before, so I'll merge this PR first although it only adds `GET` method. I think to complete the hole http library requires several PRs.

st0012:
I'll open an issue to list what's lack for the http lib, cc @adlerhsieh @janczer

PR #120, opened 2017-05-27
Fix File.chmod & Replace panic

adlerhsieh:
@janczer I found in `File.chmod` the loop does not update the last file in the variable array. Remove the `-1` to `i` index fix the problem. Could you confirm if I'm correct?

Also I believe we're replacing the built-in `panic` with `vm.returnError` so that Go's error message would not be displayed in Goby. (probably rename it to `vm.ReturnError` in the future since it's a public method.)

And for aligning the variable name, changing `v` to `vm` in `File` class since the latter is used in all other methods.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -270,7 +270,7 @@ var builtinArrayInstanceMethods = []*BuiltInMethod{
 				arr := receiver.(*ArrayObject)
 
 				if blockFrame == nil {
-					panic("Can't yield without a block")
+					vm.returnError("Can't yield without a block")
 				}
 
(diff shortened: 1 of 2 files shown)


janczer:
Yes, I think rename `vm.returnError` to `vm.ReturnError` it's good idea.

st0012:
@adlerhsieh @janczer Well I don't think export `vm.returnError` is a good idea, because it is used for reporting vm's internal error. And since it's target are vm's internal errors, other packages shouldn't have any chance to use it, so it's not necessary to export it.
Otherwise, this looks good to me too, thanks!

PR #121, opened 2017-05-28
Add Syntax Error to Parser

adlerhsieh:
Currently the parser is stuck in an endless loop when a `class` or `if` statement does not end with an `end` like:

```ruby
class Foo

# end
```

This fix allows parser to create a "syntax error" message and return the block statement immediately.

screen shot 2017-05-28 at 11 32 01 pm

If we come up with an error handling mechanism later, we can integrate this into that.


st0012:
Nice 😎

PR #131, opened 2017-06-01
Add fsm

shes50103:
Add FSM in lexer.go
Identify token type of "class" and "module"
modify parseStatement function in statement_parsing.go

### DIFF
--- a/lexer/lexer.go
+++ b/lexer/lexer.go
@@ -23,10 +23,9 @@ func New(input string) *Lexer {
 		"initial",
 		fsm.Events{
 			{Name: "method", Src: []string{"initial"}, Dst: "method"},
-			{Name: "initialize", Src: []string{"method","initial"}, Dst: "initial"},
+			{Name: "initialize", Src: []string{"method", "initial"}, Dst: "initial"},
 		},
-		fsm.Callbacks{
-		},
+		fsm.Callbacks{},
 	)
 	return l
 }
@@ -126,10 +125,10 @@ func (l *Lexer) NextToken() token.Token {
 	case '.':
 		tok = newToken(token.Dot, l.ch, l.line)
 		l.FSM.Event("method")
-		 //err := l.FSM.Event("method")
-		 //if err != nil {
-		 //    fmt.Println(err)
-		 //}
+		//err := l.FSM.Event("method")
+		//if err != nil {
+		//    fmt.Println(err)
+		//}
 	case ':':
 		if l.peekChar() == ':' {
 			l.readChar()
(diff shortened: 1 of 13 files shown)


lexer/lexer.go
```diff
@@ -2,6 +2,7 @@ package lexer
 
 import (
 	"github.com/goby-lang/goby/token"
+	"github.com/looplab/fsm"
```
st0012: You should also use `godep` to install this package into the repo.

lexer/lexer_test.go
```diff
@@ -305,16 +311,14 @@ func TestNextToken(t *testing.T) {
 		tok := l.NextToken()
 
 		if tok.Type != tt.expectedType {
-			t.Fatalf("tests[%d] - tokentype wrong. exprected=%q, got=%q", i, tt.expectedType, tok.Type)
-
+			t.Fatalf("tests[%d] - tokentype wrong. expected=%q, got=%q", i, tt.expectedType, tok.Type)
 		}
 		if tok.Literal != tt.expectedLiteral {
-			t.Fatalf("tests[%d] - literal wrong. exprected=%q, got=%q", i, tt.expectedLiteral, tok.Literal)
+			t.Fatalf("tests[%d] - literal wrong. expected=%q, got=%q", i, tt.expectedLiteral, tok.Literal)
```
st0012: Thanks for also fixed typo 😁

parser/statement_parsing.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func (p *Parser) parseStatement() ast.Statement {
 	switch p.curToken.Type {
-	case token.InstanceVariable, token.Ident, token.Constant:
+	case token.InstanceVariable, token.Ident, token.Constant, token.Class, token.Module:
```
st0012: Since we already have class token and module token, we should move them into their own cases like:
```go
switch p.curToken.Type {
case token.Class:
  p.parseClassStatement()
case token.Module:
  p.parseModuleStatement()
....
}

File: parser/statement_parsing.go [hunk 4]
```diff
@@ -8,7 +8,7 @@ import (
...
 		if p.curToken.Literal == "class" {
```
st0012: This and the module condition should be removed from this case.

lexer/lexer.go
shes50103: ok later i'll try it !

parser/parser.go
```diff
@@ -51,6 +51,8 @@ func New(l *lexer.Lexer) *Parser {
 
 	p.prefixParseFns = make(map[token.Type]prefixParseFn)
 	p.registerPrefix(token.Ident, p.parseIdentifier)
+	p.registerPrefix(token.Class, p.parseIdentifier)
```
st0012: I think this is unnecessary? Because if `class` or `module` are  used as method call, they should be classified as  `Ident` when tokenizing.

shes50103: oh yes that is unnecessary I will remove it

st0012:
@shes50103 Is this ready to merge?

Issue #129, opened 2017-06-01
Variable Assignment Does Not Return Anything


adlerhsieh:
Something like:

```ruby
a = 10
```

does not return `10` like in Ruby. It only returns if:

```ruby
a = 10
a # => 10
```

Might be a VM issue.

PR #132, opened 2017-06-02
Assign expression

st0012:
Variable assignment should be a kind of expression instead of statement, this is a prerequisite for #129


st0012:
@adlerhsieh If you reviewed and think it's ok please help me merge it.

adlerhsieh:
Looks good

Issue #125, opened 2017-05-28
Add tests for vm's error conditions


st0012:
Currently our coverage is stuck around 80% because out tests don't care about error conditions. So I think we should add tests for those conditions and remove some really edge error cases (which I add earlier for easier debugging).

st0012:
Any of you want to help me doing this? @adlerhsieh @janczer

adlerhsieh:
Currently we're using `panic` to indicate an error, which would force the program to exit. This issue might include to replace that `panic` with our builtin type `Error`, so that it is test-able. I'll take a look if I got time.

st0012:
@adlerhsieh I think [this](https://stackoverflow.com/questions/31595791/how-to-test-panics) is a good reference for this issue

adlerhsieh:
Thanks, I'll take a look.

st0012:
I think we already largely increased error condition's test coverage, so this can be closed.

PR #133, opened 2017-06-02
(WIP) Add Error Test

adlerhsieh:
Part of #125. Add basic test for "undefined method". I just test the returned object instead of `panic`.

Also, I hide error message from instruction. Or do you think it should be displayed for now for development purpose?

### DIFF
--- a/vm/error.go
+++ b/vm/error.go
@@ -9,6 +9,8 @@ var (
 	UndefinedMethodErrorClass *RClass
 	// ArgumentErrorClass ...
 	ArgumentErrorClass *RClass
+	// TypeErrorClass ...
+	TypeErrorClass *RClass
 )
 
 func initErrors() {
@@ -16,6 +18,8 @@ func initErrors() {
 	UndefinedMethodErrorClass = &RClass{BaseClass: bc}
 	bc = createBaseClass("ArgumentError")
 	ArgumentErrorClass = &RClass{BaseClass: bc}
+	bc = createBaseClass("TypeError")
+	TypeErrorClass = &RClass{BaseClass: bc}
 }
 
 // Error ...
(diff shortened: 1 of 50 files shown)


st0012:
@adlerhsieh This reminds me that we should start developing error classes (like `StandardError`), then we can compare test cases using the class name (it will be like the `errorType` you wrote) and the message it contains. What do you think?

adlerhsieh:
I agree that sooner or later we need to do that. I'll try one or two in this PR And see how it is like.

vm/vm.go
```diff
@@ -171,7 +171,7 @@ func (vm *VM) execInstruction(cf *callFrame, i *instruction) {
 			if stackTrace == 0 {
 				fmt.Printf("Internal Error: %s\n", p)
 			}
-			fmt.Printf("Instruction trace: %d. \"%s\"\n", stackTrace, i.inspect())
+			// fmt.Printf("Instruction trace: %d. \"%s\"\n", stackTrace, i.inspect())
```
st0012: I think comment out this is ok. This only works when running single test case.

adlerhsieh:
@st0012 Could you check in [this commit](https://github.com/goby-lang/goby/pull/133/commits/ff7a098b7f6076d60ea16ef78a7ebd04cf79038a) if this is what you expect? I've added a `ArgumentErrorObject` as an example so that code with this error will return an error object. 

If you feel it's ok then I'll start implement other errors. Probably after this we can move the `panic` to a point where the vm checks if a returned object is an error. If so, start handling this error.

vm/error.go
```diff
@@ -1,18 +1,68 @@
...
+// RError ...
+type RError struct {
+  *BaseClass
 }
 
+// Error ...
 type Error struct {
-	Class   *ErrorClass
-	Message string
+  Class   *RError
```
st0012: Actually I'm thinking using `*RClass` is fine. Because we may not do any performance optimization on errors so we don't need a special class type for it. Besides using `*RClass` means users can create inherit their own exception classes like in Ruby.

adlerhsieh:
The commits in this PR is a little messy. I'll create another one.

Issue #119, opened 2017-05-27
Question: class can inherit module?


hachi8833:
I have a small question.

I found you can inherit modules to class as if modules were a class.

The following works on the latest Goby build, but I just wonder if this behavior is OK.

```ruby
module Foo
  def self.my
    999
  end
  def ten
    100
  end
  def five
     7
  end
end

class Bar < Foo
  def five
    5
  end
end

a = Bar.new
puts(a.ten * a.five ) # => 500
puts(Bar::Foo.my)  # => 999
```

If you run the code with Ruby, you receive "superclass must be a Class (Module given) (TypeError)."
If you change `module` to `class` in line 1, it works with Ruby.

hachi8833:
One more question.
Are the embedded classes closed?

Looks like the embedded classes are not extensible so far: I cannot add or override methods against them, and cannot inherit them. 

```ruby
class String
  def my_method
    self + "47"
  end
end

a = "test"
puts(a.my_method)  # => undefined method `my_method' for 1
```

```ruby
class MyString < String
  def my_method
    self + "47"
  end
end


i = MyString.new("string")
puts(i.my_method)  # => Constant String is not a class. got=CLASS
```

I'm just unsure if this behavior is the expected one, or unimplemented. This depends on your plan or specification.

adlerhsieh:
I don't think we have implemented self-defined method for built-in class, so defining extra methods  for `String` is not working. And yes, from a design perspective module should not be inherited to a class. It would be great to implement those features if you like to try :)

adlerhsieh:
Unless @st0012 got a different idea on this

hachi8833:
Thank you for the answer!

FYI: now I found that built-in class can be extended and overload methods by including a module:

```ruby
module MyModule
  def my_method
    "47"
  end

  def to_i
    48
  end
end

class String
  include(MyModule)
end


i = String.new("string")
puts(i.my_method)  # => 47
puts(i.to_i)       # => 48
```

In Ruby, `puts(i.to_i)` returns `0`.

---------------------

Just a mutter to myself: Module's behavior is a little bit different between current Goby and Ruby, but this implies that built-in classes could be extended with modules that are written in Goby. 

(Actually, I've been wondering if built-in classes such as Array and Hash can be extended with Enumerable module written in Goby, which might make Goby extensible more easily. To do this, perhaps I need Enumerator class first :-p. )

st0012:
[...]

PR #136, opened 2017-06-04
Prohibit module inheritance

hachi8833:
(see #119 )
Sorry for the delay. I'd be glad if some of you'd take a look at the fix. 

Test code for the fix is still needed. 
I'd add the test when #133 is merged to master.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -404,8 +404,12 @@ var builtinClassClassMethods = []*BuiltInMethodObject{
 		Name: "new",
 		Fn: func(receiver Object) builtinMethodBody {
 			return func(vm *VM, args []Object, blockFrame *callFrame) Object {
-
 				class := receiver.(*RClass)
+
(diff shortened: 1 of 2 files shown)


hachi8833:
I'm unsure how to add a flag to codecov coverage...

adlerhsieh:
Could you also add a test for this case? The reason why codecov does not pass is because we add some code without giving a test case (which decrease the % of coverage).

You can add a case in `vm/class_test.go` like:

```go
func TestClassInheritModule(t *testing.T) {
  input := `
module Foo
end

class Bar < Foo
end
`

  evaluated := testEval(t, input)
  // check if `evaluated` is an error or contains error message
}
```

You can find several examples for testing errors in the same file. 

Make sure it pass and push again :)

st0012:
LGTM
@adlerhsieh You can merge it once you reviewed.

adlerhsieh:
Looks good. Thanks!

PR #142, opened 2017-06-06
Thread on simple server

st0012:
## Load testing
Use [Boom](https://github.com/tarekziade/boom) as load testing tool. Apache Bench sometimes just hangs there.
Test script I use:
```
$ boom http://localhost:3000/ -n 100000 -c 100
```

And I call sleep function in:

```go
http.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
    thread := t.vm.newThread()
    if t.vm.threadCount % 100 == 0 {
        log.Println("-------------------Start Sleeping-------------------------")
        time.Sleep(10 * time.Second)
        log.Println("###################Stop Sleeping########################")
    }
    res := httpResponseClass.initializeInstance()
    ......
}
```

for testing if it's really running concurrently or one slow request blocks others.

## Profiling
Use [gcvis](https://github.com/davecheney/gcvis) for monitoring memory usage. Run
```
$ env GOMAXPROCS=4 gcvis go run goby.go samples/server.gb 
```

## Conclusion

- It turns out create a new thread on every requests works pretty well (at least for now). 
- The final `i` in `server.gb` will match the total request number, which means there's no obvious race condition 😁.
- The blocking test's result is as I expected: just the selected threads sleeps (and will recover later) and we can still serve other requests at the same time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: go
 go:
   - 1.8.x
 
+before_install:
+  - sudo apt-get install apache2-utils
+
 script:
   - make test
 
--- a/bzt.yml
+++ b/bzt.yml
@@ -1,0 +1,14 @@
+# Please install Taurus first
+# Then run `bzt bzt.yml
+# See https://gettaurus.org/docs/ConfigSyntax/ for more setting options
+execution:
+  - concurrency: 50
+    throughput: 10 # requests per second
+    ramp-up: 10s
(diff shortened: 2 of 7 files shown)


st0012:
@adlerhsieh @hachi8833 I'll merge this first because current server is fragile, but I still recommend you to test it yourself to see if I miss anything.

PR #144, opened 2017-06-07
Add New Error Types

adlerhsieh:
This PR adds the following error classes:

- UndefinedMethodError
- TypeError
- ArgumentError

Also, in `vm/instructions.go`, a new way of returning error is added. If an error is occurred when running `evalCallFrame`, it will return and stop the thread.

The existing way of `panic` is still there for most cases. If this PR is ok, then I'll start replacing all existing `panic` with this type of error handling.

### DIFF
--- a/vm/array_test.go
+++ b/vm/array_test.go
@@ -425,13 +425,13 @@ func TestCountMethodFail(t *testing.T) {
 	for _, tt := range testsFail {
 		evaluated := testEval(t, tt.input)
 
-		err, ok := evaluated.(*ArgumentErrorObject)
+		err, ok := evaluated.(*Error)
 		if !ok {
 			t.Errorf("Expect ArgumentError. got=%T (%+v)", err, err)
 		}
 
-		if err.Message != tt.expected.Message {
-			t.Errorf("Expect error message \"%s\". got=\"%s\"", tt.expected.Message, err.Message)
+		if err.Class.ReturnName() != ArgumentError {
+			t.Errorf("Expect \"%s\". got=\"%s\"", ArgumentError, err.Class.ReturnName())
 		}
 	}
 }
--- a/vm/array_test.go
+++ b/vm/array_test.go
@@ -426,12 +426,8 @@ func TestCountMethodFail(t *testing.T) {
 		evaluated := testEval(t, tt.input)
 
 		err, ok := evaluated.(*Error)
-		if !ok {
+		if !ok || err.Class.ReturnName() != ArgumentError {
 			t.Errorf("Expect ArgumentError. got=%T (%+v)", err, err)
-		}
-
-		if err.Class.ReturnName() != ArgumentError {
-			t.Errorf("Expect \"%s\". got=\"%s\"", ArgumentError, err.Class.ReturnName())
 		}
 	}
 }
--- a/vm/array_test.go
+++ b/vm/array_test.go
@@ -424,10 +424,12 @@ func TestCountMethodFail(t *testing.T) {
 
 	for _, tt := range testsFail {
 		evaluated := testEval(t, tt.input)
-		err, ok := evaluated.(*Error)
+
+		err, ok := evaluated.(*ArgumentErrorObject)
 		if !ok {
-			t.Errorf("Expect error. got=%T (%+v)", err, err)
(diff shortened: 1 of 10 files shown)


vm/error.go
```diff
@@ -1,18 +1,95 @@
...
+func (e *UndefinedMethodErrorObject) Inspect() string {
+	return "ArgumentError: " + e.Message
+}
+
+func (e *UndefinedMethodErrorObject) returnClass() Class {
+	return e.Class
+}
+
+func initializeUndefinedMethodError(format string, args ...interface{}) *UndefinedMethodErrorObject {
+	return &UndefinedMethodErrorObject{Class: UndefinedMethodErrorClass, Message: fmt.Sprintf(format, args...)}
```
st0012: Why do we need to store message in another struct instead of Basic Error struct? like

```go
return &Error{Class: UndefinedMethodErrorClass, Message: fmt.Sprintf(format, args...)}
```

Then we don't have to create struct for every type of exceptions

vm/error.go
```diff
@@ -1,18 +1,95 @@
...
+	// UndefinedMethodErrorClass ...
+	UndefinedMethodErrorClass *RClass
+	// ArgumentErrorClass ...
+	ArgumentErrorClass *RClass
+	// TypeErrorClass ...
+	TypeErrorClass *RClass
+)
+
+func initErrors() {
+	bc := createBaseClass("UndefinedMethodError")
```
st0012: You can set these exception name into constants like

```go
const(
    UndefinedMethodError = "UndefinedMethodError"
)
```

So that we can use this constants in other places instead of copy and paste strings.

vm/error_test.go
```diff
@@ -0,0 +1,35 @@
...
+	_, ok := evaluated.(*UndefinedMethodErrorObject)
+	if !ok {
+		t.Errorf("Expect UndefinedMethodError. got=%T (%+v)", evaluated, evaluated)
+	}
+}
+
+func TestUndefinedMethodErrorStopsProgram(t *testing.T) {
+	evaluated := testEval(t, `a; 1 + 1`)
+	_, ok := evaluated.(*UndefinedMethodErrorObject)
+	if !ok {
```
st0012: If you do what I said above, you can change this into
```go
errObj, ok := evaluated.(*Error)
if !ok {
  // Expect object to be error object
}

if errObj.class.ReturnName != UndefinedMethodError // This is constant {
    // And this can be wrapped as a function
    t.Errorf("Expect %s. got=%T (%+v)", UndefinedMethodError, evaluated, evaluated)
}

**Comment by adlerhsieh** [comment]
I've made some changes accordingly. You're right. It does save us from creating all types of errors.

File: vm/array_test.go [hunk 4]
```diff
@@ -424,12 +424,14 @@ func TestCountMethodFail(t *testing.T) {
 
 	for _, tt := range testsFail {
 		evaluated := testEval(t, tt.input)
+
 		err, ok := evaluated.(*Error)
 		if !ok {
```
st0012: I think this can be shorten like:
```go
if !ok || err.Class.ReturnName() != ArgumentError {
    t.Errorf("Expect ArgumentError. got=%T (%+v)", err, err)
}

Since we basically return similar fail message, this will be simpler to write.
However, this is just a thought not change request.

File: vm/error.go [hunk 5]
```diff
@@ -1,18 +1,68 @@
...
+// Inspect ...
 func (e *Error) Inspect() string {
 	return "ERROR: " + e.Message
 }
 
 func (e *Error) returnClass() Class {
 	return e.Class
 }
+
+func initializeUndefinedMethodError(format string, args ...interface{}) *Error {
```
st0012: Since we don't want to create struct for every type of error, I think we also shouldn't create a function for every error type.

I would write something like:

```go
func initializeError(errorType *RClass, format string, args ...interface{}) *Error {
    return &Error{
        Class: errorType,
        Message: fmt.Sprintf(errorType.class.Name + ": "+format, args...),
    }
}

And this can be used on any types of error.

adlerhsieh: That's a good point. Thanks.

adlerhsieh:
Updated.

st0012:
@adlerhsieh Ok, just rebase it and lets merge and give it a try 😄

PR #146, opened 2017-06-08
WIP:Method call without parens v2

shes50103:
- Support function definition without parens.
- Support method call without parens.
- Support parameter of method call in type Ident, Int, InstanceVariable, String, Boolean
Now we can write goby like this!
`       

	  def set_x x1
	    @x = x1
	  end
	
	  def set_y y1, y2, y3
	    @y = y1 + y2 + y3
	  end
	
	  f = Foo.new
	  f.set_x 1
	  f.set_y 4,5,6
	  f.x + f.y

`

If you find some code is not appropriate just leave a message and let me know~

### DIFF
--- a/parser/expression_parsing.go
+++ b/parser/expression_parsing.go
@@ -65,8 +65,7 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 	for !p.peekTokenIs(token.Semicolon) && precedence < p.peekPrecedence() && p.peekTokenAtSameLine() {
 
 		infix := p.infixParseFns[p.peekToken.Type]
-		if p.peekTokenIs(token.Int) && precedence == 0 {
-			// method call without paren case
+		if (p.peekTokenIs(token.Int) || p.peekTokenIs(token.Ident) || p.peekTokenIs(token.InstanceVariable)) && precedence == 0 { // method call without paren case
 			infix = p.parseCallExpression
 		}
 
@@ -310,8 +309,7 @@ func (p *Parser) parseIfExpression() ast.Expression {
 func (p *Parser) parseCallExpression(receiver ast.Expression) ast.Expression {
 	var exp *ast.CallExpression
 
-	if p.curTokenIs(token.LParen) { // call expression doesn't have a receiver foo(x) || foo()
-		// method name is receiver, for example 'foo' of foo(x)
+	if p.curTokenIs(token.LParen) || p.curTokenIs(token.Ident) || p.curTokenIs(token.Int) || p.curTokenIs(token.InstanceVariable) {
 
 		m := receiver.(*ast.Identifier).Value
 		// receiver is self
@@ -321,18 +319,13 @@ func (p *Parser) parseCallExpression(receiver ast.Expression) ast.Expression {
 
 		// current token is identifier (method name)
 		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver, Method: m}
-		exp.Arguments = p.parseCallArguments()
-	} else if p.curTokenIs(token.Ident) || p.curTokenIs(token.Int) {
 
-		m := receiver.(*ast.Identifier).Value
-		// receiver is self
-		selfTok := token.Token{Type: token.Self, Literal: "self", Line: p.curToken.Line}
-		self := &ast.SelfExpression{Token: selfTok}
-		receiver = self
+		if p.curTokenIs(token.LParen) {
+			exp.Arguments = p.parseCallArguments()
 
-		// current token is identifier (method name)
-		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver, Method: m}
-		exp.Arguments = p.parseCallArgumentsWithoutParens()
+		} else {
+			exp.Arguments = p.parseCallArgumentsWithoutParens()
+		}
 
 	} else if p.curTokenIs(token.Dot) { // call expression has a receiver like: p.foo
 
--- a/parser/expression_parsing.go
+++ b/parser/expression_parsing.go
@@ -314,7 +314,18 @@ func (p *Parser) parseCallExpression(receiver ast.Expression) ast.Expression {
 		// current token is identifier (method name)
 		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver, Method: m}
 		exp.Arguments = p.parseCallArguments()
-	} else { // call expression has a receiver like: p.foo
+	} else if p.curTokenIs(token.Ident){
+		m := receiver.(*ast.Identifier).Value
+		// receiver is self
+		selfTok := token.Token{Type: token.Self, Literal: "self", Line: p.curToken.Line}
+		self := &ast.SelfExpression{Token: selfTok}
+		receiver = self
+
+		// current token is identifier (method name)
+		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver, Method: m}
+		exp.Arguments = p.parseCallArgumentsWithoutParens()
+
+	}else if p.curTokenIs(token.Dot) { // call expression has a receiver like: p.foo
 		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver}
 
 		// check if method name is identifier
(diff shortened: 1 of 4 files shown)


vm/evaluation_test.go
```diff
@@ -786,3 +786,60 @@ func TestMethodCallWithNestedBlock(t *testing.T) {
...
+		{
+			`
+			class Foo
+			  def set_x x1, x2, x3
+			    @x = x1
+			    @y = x2
+			  end
+
+			  def foo
+			    set_x(10,11,12)
```
st0012: Actually you can use `attr_reader` here, then you can remove this method and just write:

```ruby
f.x + a + f.y

File: parser/expression_parsing.go [hunk 2]
```diff
@@ -313,7 +314,20 @@ func (p *Parser) parseCallExpression(receiver ast.Expression) ast.Expression {
 		// current token is identifier (method name)
 		exp = &ast.CallExpression{Token: p.curToken, Receiver: receiver, Method: m}
 		exp.Arguments = p.parseCallArguments()
-	} else { // call expression has a receiver like: p.foo
+	} else if p.curTokenIs(token.Ident) {
+
```
st0012: Make sure you refactor this part and above section since they have only a few differences

st0012:
@shes50103 What blocks you from removing parens on method call? Do you think this requires redesign parser?

shes50103:
@st0012 The bottleneck  is I don't out how to set precedence in parens on method call.
I think I can figure it out after I get more familiar with this code!

parser/expression_parsing.go
```diff
@@ -59,10 +59,16 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 		p.noPrefixParseFnError(p.curToken.Type)
 		return nil
 	}
+
 	leftExp := prefix()
 
 	for !p.peekTokenIs(token.Semicolon) && precedence < p.peekPrecedence() && p.peekTokenAtSameLine() {
+
 		infix := p.infixParseFns[p.peekToken.Type]
+		if (p.peekTokenIs(token.Int) || p.peekTokenIs(token.Ident) || p.peekTokenIs(token.InstanceVariable)) && precedence == 0 { // method call without paren case
```
st0012: What about string argument or boolean argument? The basic idea is right, but the implementation needs more work. I think you should create a hash table that contains types of token that can be consider as argument.

parser/expression_parsing.go
```diff
@@ -59,10 +59,16 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 		p.noPrefixParseFnError(p.curToken.Type)
 		return nil
 	}
+
 	leftExp := prefix()
 
 	for !p.peekTokenIs(token.Semicolon) && precedence < p.peekPrecedence() && p.peekTokenAtSameLine() {
+
 		infix := p.infixParseFns[p.peekToken.Type]
```
st0012: This should be the `else` condition of the below one. It's like more a fallback now.

1. We look at next token to see if it is a call expression without paren.
2. If not, we see if it is an infix expression.
3. It's not in above conditions, just a single expression.

parser/expression_parsing.go
```diff
@@ -303,8 +309,8 @@ func (p *Parser) parseIfExpression() ast.Expression {
 func (p *Parser) parseCallExpression(receiver ast.Expression) ast.Expression {
 	var exp *ast.CallExpression
 
-	if p.curTokenIs(token.LParen) { // call expression doesn't have a receiver foo(x) || foo()
-		// method name is receiver, for example 'foo' of foo(x)
+	if p.curTokenIs(token.LParen) || p.curTokenIs(token.Ident) || p.curTokenIs(token.Int) || p.curTokenIs(token.InstanceVariable) {
```
st0012: Why do you repeat same logic again here?

parser/statement_parsing.go
```diff
@@ -63,11 +64,15 @@ func (p *Parser) parseDefMethodStatement() *ast.DefStatement {
 	}
 	// def foo
 	if p.peekTokenAtSameLine() { // def foo(), next token is ( and at same line
-		if !p.expectPeek(token.LParen) {
-			return nil
+		if p.peekTokenIs(token.LParen) {
+			if !p.expectPeek(token.LParen) {
```
st0012: If you confirmed next token is `LParen` above, why check it again here? This will always be false.

parser/statement_parsing.go
```diff
@@ -63,11 +64,15 @@ func (p *Parser) parseDefMethodStatement() *ast.DefStatement {
...
+				return nil
+			}
+			stmt.Parameters = p.parseParameters()
+		} else if p.peekTokenIs(token.Ident) {
```
st0012: Please add an example in comment.

parser/statement_parsing.go
```diff
@@ -170,7 +193,11 @@ func (p *Parser) parseReturnStatement() *ast.ReturnStatement {
 func (p *Parser) parseExpressionStatement() *ast.ExpressionStatement {
 	stmt := &ast.ExpressionStatement{Token: p.curToken}
 
-	stmt.Expression = p.parseExpression(LOWEST)
+	precedence := 1
```
st0012: You shouldn't use int to express precedence directly, use constants. For example, precedence 1 is `LOGIC`.

st0012: BTW please also add comments for this change (include below).

vm/evaluation_test.go
```diff
@@ -786,3 +786,138 @@ func TestMethodCallWithNestedBlock(t *testing.T) {
...
+			f = Foo.new
+			f.foo
+			`,
+			31,
+		},
+		{
+			`
+			class Foo
+			  def set_x x1, x2, x3
+			    @x1 = x1
```
st0012: I think these test cases are a little unnecessarily verbose. Just write something like:

```ruby
class Foo
  attr_reader("x", "y")

  def set_x x1, x2, x3
    @x = x1 + x2 + x3
  end

  def set_y y1, y2, y3
    @y = y1 + y2 + y3
  end
end

f = Foo.new
f.x + f.y
```

st0012: Same rule applies to other test inputs

st0012:
@shes50103 BTW you should also put some before/after samples on description, this is a huge improvement 😁

parser/expression_parsing.go
shes50103: OK I just create a  hash table which include  string argument, boolean argument and others.

parser/statement_parsing.go
shes50103: ok I will simplify it

parser/statement_parsing.go
shes50103: @st0012 
 I will add a new constants to express precedence which is call_without_parens case.
But I just name new constants "FIRST" which is lower than "LOWEST" constant.
Is it appropriate to name it "FIRST" before "LOWER"?
or I just change current "LOWEST" to be "LOWER" or something
and make my new constants be "LOWEST"?

st0012: `LOWEST` should be literally the lowest.
I think a better way is to name your constant `CALL_WITHOUT_PAREN`, which is very verbose but also clear. And rename `LOWEST` to `NORMAL`.

vm/evaluation_test.go
shes50103: I will remain some test and add the case you mention!

st0012: ok

parser/expression_parsing.go
```diff
@@ -8,6 +8,15 @@ import (
 	"github.com/goby-lang/goby/token"
 )
 
+var argument = map[token.Type]bool{
```
st0012: `arguments`

parser/expression_parsing.go
```diff
@@ -36,6 +45,7 @@ var precedence = map[token.Type]int{
 // Constants for denoting precedence
 const (
 	_ int = iota
+	FIRST
```
st0012: This should be `LAST`, but I'll change these constant names myself.

parser/expression_parsing.go
st0012: But I'll change it

PR #155, opened 2017-06-10
Add API documents to Class

hachi8833:
Added API documents to methods in Class class.

This depends upon my understanding of Goby as well as Ruby.
Especially `@params` and `@return` needs check.

Thank you,

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -418,7 +418,7 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
 		// puts("123".class.name)  # => String
 		// ```
 		//
-		// @params object [Object] Receiver (required)
+		// @param object [Object] Receiver (required)
 		// @return [Class] The class of the receiver
 		Name: "class",
 		Fn: func(receiver Object) builtinMethodBody {
@@ -441,7 +441,7 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
 		// !false # => true
 		// ```
 		//
-		// @params object [Object] object that return boolean value to invert
+		// @param object [Object] object that return boolean value to invert
 		// @return [Object] Inverted boolean value
 		Name: "!",
 		Fn: func(receiver Object) builtinMethodBody {
@@ -461,7 +461,7 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
 		// puts(a)     # => 2
 		// ```
 		//
-		// @params sec [Integer] time to wait in sec
+		// @param sec [Integer] time to wait in sec
 		// @return [Integer] actual time slept in sec
 		Name: "sleep",
 		Fn: func(receiver Object) builtinMethodBody {
@@ -492,7 +492,7 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
 		// end
 		// ```
 		//
-		// @params n/a []
+		// @param n/a []
 		// @return [Boolean] true/false
 		Name: "block_given",
 		Fn: func(receiver Object) builtinMethodBody {
@@ -612,7 +612,7 @@ var builtinClassClassMethods = []*BuiltInMethodObject{
 		// end
 		// ```
 		//
-		// @param *args [String] One or more quoted method names for 'getter/getter'
+		// @param *args [String] One or more quoted method names for 'getter/setter'
 		// @return [Null]
 		Name: "attr_accessor",
 		Fn: func(receiver Object) builtinMethodBody {
--- a/vm/class.go
+++ b/vm/class.go
@@ -22,8 +22,8 @@ func initTopLevelClasses() {
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


hachi8833:
Thank you for looking into it! I'll fix them.

hachi8833:
Sorry, I should've waited for finishing.
Rephrasing is OK.

adlerhsieh:
The last thing I would like to discuss is the style of documentation in Go. In most cases we do comments with a leading space between `//` and the actual comments. For example:

```go
// This is a comment
```

Instead of:

```go
//This is a comment
```

I know both ways work and there is no convention or rules saying that we should do in which way. However, I think it is better to follow the community and use the first approach. If you agree with me, could you add a space to all the comments in this file? We can discuss if you have another idea, thanks!

st0012:
@adlerhsieh @hachi8833 You guys can merge this PR once you reach a consensus.

vm/class.go
```diff
@@ -353,6 +434,15 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
...
 	},
 	{
+		// Inverts the boolean value.
+		//
+		// ```ruby
+		// !true  # => false
+		// !false # => true
+		// ```
+		//
+		// @params object [Object] object that return boolean value to invert
```
adlerhsieh: This should be `@param` instead of `@params`

vm/class.go
```diff
@@ -339,6 +407,19 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
...
+		// FYI: You can convert the class into String with `#name`.
+		//
+		// ```ruby
+		// puts(100.class)         # => <Class:Integer>
+		// puts(100.class.name)    # => Integer
+		// puts("123".class)       # => <Class:String>
+		// puts("123".class.name)  # => String
+		// ```
+		//
+		// @params object [Object] Receiver (required)
```
adlerhsieh: This should be `@param` instead of `@params`

vm/class.go
```diff
@@ -287,6 +305,28 @@ func createBaseClass(className string) *BaseClass {
...
+		// Currently, only the following embedded Goby libraries are targeted:
+		//
+		// - "file"
+		//
+		// - "net/http"
+		//
+		// - "net/simple_server"
+		//
+		// - "uri"
+		//
```
adlerhsieh: Sorry I didn't make it very clear. The blank line is only for separating bullet points and a ordinary content, not for each point. So it should be:

```go
// Currently, only the following embedded Goby libraries are targeted:
//
// - point
// - point
// - point
// - point
```

vm/class.go
```diff
@@ -362,6 +452,17 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
...
+		// Suspends the current thread for duration (sec).
+		//
+		// **Note:** currently, parameter cannot be omitted, and only Integer can be specified.
+		//
+		// ```ruby
+		// a = sleep(2)
+		// puts(a)     # => 2
+		// ```
+		//
+		// @params sec [Integer] time to wait in sec
```
adlerhsieh: This should be `@param` instead of `@params`

vm/class.go
```diff
@@ -373,6 +474,26 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
...
+		//     if block_given
+		//       yield(file)
+		//     end
+		//
+		//     file.close
+		//   end
+		// end
+		// ```
+		//
+		// @params n/a []
```
adlerhsieh: This should be `@param` instead of `@params`

vm/class.go
```diff
@@ -413,6 +584,36 @@ var builtinClassClassMethods = []*BuiltInMethodObject{
...
+		//   def bar=(val)
+		//     @bar = val
+		//   end
+		//   def buz=(val)
+		//     @buz = val
+		//   end
+		// end
+		// ```
+		//
+		// @param *args [String] One or more quoted method names for 'getter/getter'
```
adlerhsieh: I suppose`'getter/getter'` should be `'getter/setter'`?

vm/class.go
hachi8833: The blank line is needed for generating HTML: they would be like `-point -point -point -
 point` if  no blank.

vm/class.go
hachi8833: Fixed all of them.

vm/class.go
hachi8833: Fixed as well.

st0012:
@hachi8833 Did you rebased the right way?

st0012:
Your commit tree is quite ugly https://github.com/goby-lang/goby/network .Please fix this

hachi8833:
Sorry, looks like the rebase in a wrong way.
I'll fix it.

hachi8833:
I'd close the PR and repush the clean branch for the fixes, referring to the PR to track the review.
I hope this works.

PR #156, opened 2017-06-11
[Retry] Add API documents to Class

hachi8833:
This is the fix for wrong rebasing on #155. Sorry. 
See https://github.com/goby-lang/goby/pull/155 for the review history.

Note that I left the following blank lines because they need just for generating HTML on API doc: they would be shown as `-point -point -point -point` on HTML if no blank lines are added.

```diff
+		// - "file"
+		//
+		// - "net/http"
+		//
+		// - "net/simple_server"
+		//
+		// - "uri"
```

### DIFF
--- a/samples/to_json.gb
+++ b/samples/to_json.gb
@@ -1,0 +1,2 @@
+h = { a: 1, b: [1, "2", [4, 5, nil], { foo: "bar" }]}.to_json
+puts(h) #=> {"a":1,"b":[1, "2", [4, 5, null], {"foo":"bar"}]}
--- a/vm/array.go
+++ b/vm/array.go
@@ -39,6 +39,20 @@ func (a *ArrayObject) Inspect() string {
 	return out.String()
 }
 
+func (a *ArrayObject) toJSON() string {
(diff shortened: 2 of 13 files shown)


adlerhsieh:
For markdown format, bullet points are treated differently when there's a blank line between each points. They are both bullet points, instead of treated as the same line. If there's a blank line between each point, there is more spacing between each point. The following is an example we can see using GitHub issue:

```markdown
Example A

- point1
- point2
- point3

Example B

- point1

- point2

- point3
```

is rendered into:

> Example A
> 
> - point1
> - point2
> - point3
> 
> Example B
> 
> - point1
> 
> - point2
> 
> - point3

The following is a screenshot I use with our current markdown parser.

screen shot 2017-06-11 at 5 24 13 pm

It works like in GitHub issue, so I think it's better to remove those blank lines between each point.

The rest looks good.

hachi8833:
Thank you for the investigation! Resolved the issue.

adlerhsieh:
Thanks! Could you rebase and git push --force so that I can merge?

PR #157, opened 2017-06-11
Remove parens

st0012:
Thanks to @shes50103 we can call methods without wrapping parens on arguments. So this PR is to remove unusual paren wrappings (like `require("foo")` or `include(Bar)` in test cases and sample files.

### DIFF
--- a/parser/expression_parsing.go
+++ b/parser/expression_parsing.go
@@ -20,11 +20,11 @@ var argument = map[token.Type]bool{
 var precedence = map[token.Type]int{
 	token.Eq:                 EQUALS,
 	token.NotEq:              EQUALS,
-	token.LT:                 LESSGREATER,
-	token.LTE:                LESSGREATER,
-	token.GT:                 LESSGREATER,
-	token.GTE:                LESSGREATER,
-	token.COMP:               LESSGREATER,
+	token.LT:                 COMPARE,
+	token.LTE:                COMPARE,
+	token.GT:                 COMPARE,
+	token.GTE:                COMPARE,
+	token.COMP:               COMPARE,
 	token.And:                LOGIC,
 	token.Or:                 LOGIC,
 	token.Plus:               SUM,
@@ -45,11 +45,11 @@ var precedence = map[token.Type]int{
 // Constants for denoting precedence
 const (
 	_ int = iota
-	FIRST
 	LOWEST
+	NORMAL
 	LOGIC
 	EQUALS
-	LESSGREATER
+	COMPARE
 	ASSIGN
 	SUM
 	PRODUCT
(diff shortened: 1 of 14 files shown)


parser/expression_parsing.go
```diff
@@ -8,23 +8,25 @@ import (
...
-var argument = map[token.Type]bool{
+var arguments = map[token.Type]bool{
 	token.Int:              true,
 	token.String:           true,
 	token.True:             true,
 	token.False:            true,
+	token.Null:             true,
 	token.InstanceVariable: true,
 	token.Ident:            true,
+	token.Constant:         true,
```
st0012: @shes50103 You forgot constants and nil can also be arguments.

st0012:
@adlerhsieh @shes50103 Please take a glance at this PR, I want to merge it ASAP, thanks

parser/expression_parsing.go
shes50103: oh! thanks to help me add it!

hachi8833:
Looks like the following cases should also be covered:

-------

```ruby
require("file") # unnecessary for Ruby
puts(File.extname("test.gb"))
```

The code above, with full parentheses, works fine. 

* Goby:
> .gb

## 1: Removing parentheses from `puts`:

```ruby
require("file") # unnecessary for Ruby
puts File.extname("test.gb")
```

* Ruby(remove `require`):
> .gb

* Goby: no outputs

## 2: Removing parentheses from `File.extname`:

```ruby
require("file") # unnecessary for Ruby
puts(File.extname "test.gb")
```

* Ruby(remove `require`):
> .gb

* Goby:
> Internal Error: runtime error: index out of range
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

## 3: Removing all parentheses

```ruby
require("file") # unnecessary for Ruby
puts File.extname "test.gb"
```

* Ruby(remove `require`):
> .gb

* Goby: 
> Internal Error: runtime error: index out of range
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

--------------

Implicit nests with multiple arguments might make things harder to be handled...

st0012:
@shes50103 You can add more tests if you're worried. And you can use `codecov` to check where's the test is missing. For example, [this method](https://codecov.io/gh/goby-lang/goby/compare/ce466ff52376ffb8fdf0f7d3bba4e40e97b84f44...d1378f40ffdd76074fa9dfe7a03c022ba06af51e/src/parser/expression_parsing.go#L441) hasn't been ran through whole test, if you missed conditions that use this method?

PR #159, opened 2017-06-11
Hash to json

st0012:
Just like `Hash#to_json` in Rails, example:
```ruby
h = { a: 1, b: [1, "2", [4, 5, nil], { foo: "bar" }]}.to_json
puts(h) #=> {"a":1,"b":[1, "2", [4, 5, null], {"foo":"bar"}]}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ It will have Ruby's syntax (I'll try to support all common syntaxes) but without
     - String
     - Boolean
     - nil
-    - Hash
+    - Hash (with built in `to_json` method)
     - Array
 - Flow control
     - If statement
--- a/vm/hash.go
+++ b/vm/hash.go
@@ -77,6 +77,14 @@ func (h *HashObject) toJSON() string {
 
 var builtinHashInstanceMethods = []*BuiltInMethodObject{
 	{
+		// Returns json that is corresponding to the hash.
+		// Basically just like Hash#to_json in Rails but currently doesn't support options.
+		//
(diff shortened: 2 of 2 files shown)


st0012:
BTW I can't use string for test comparison because Go's map (which we use to store key/value pairs) is random ordered. So we might get different result string every time.

hachi8833:
Got an error with the following that contains capitalized key(yes, not so good):

```ruby
h = { Ocean:  { Squid: 10, Octopus: 8 }}.to_json
puts(h)
```

> panic: expected next token to be IDENT, got CONSTANT instead. Line: 0
expected next token to be }, got CONSTANT instead. Line: 0
no prefix function for :. Line: 0
expected next token to be IDENT, got CONSTANT instead. Line: 0
expected next token to be }, got CONSTANT instead. Line: 0
no prefix function for :. Line: 0
no prefix function for ,. Line: 0
no prefix function for :. Line: 0
no prefix function for }. Line: 0
no prefix function for }. Line: 0
no prefix function for .. Line: 0

But the following works:

```ruby
ruby
h = { oCean:  { sqUid: 10, ocTopus: 8 }}.to_json
puts(h)
```

st0012:
@hachi8833 well that's not `to_json`'s issue. It's because we haven't support constants as hash's key,  can you open an issue for this and I or @hachi8833  will open another PR to deal with it.

st0012:
If there's no other issues, I'm going to merge this~

PR #162, opened 2017-06-14
Symbol

shes50103:
- Support Symbol token in lexer
- Identify ":" is a mark of symbol or just a Colon

Following is support syntax:

```
:apple
{ test:"abc" }
{ test: :abc }
{ test: 50 }
{ test: abc }
{ test:abc }
```
Not  support this syntax:
```
{ test::abc }
```
Now you can write Goby like this
```
attr_reader :x, :y
```

### DIFF
--- a/lexer/lexer.go
+++ b/lexer/lexer.go
@@ -3,6 +3,7 @@ package lexer
 import (
 	"github.com/goby-lang/goby/token"
 	"github.com/looplab/fsm"
+	"fmt"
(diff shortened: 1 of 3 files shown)


st0012:
@shes50103 Looks like it's finished?

shes50103:
Yes, it's finished!

st0012:
@shes50103 Thanks!

PR #167, opened 2017-06-18
Add sleep to test.sh to avoid 'connection refused' error on apache bench

hachi8833:
Just to avoid the following error that occurs occasionally on local `make test`:

> Benchmarking localhost (be patient)
apr_socket_recv: Connection refused (61)
make: *** [test] Error 61
2017/06/18 18:34:16 SimpleServer start listening on port: 3000
^C2017/06/18 18:34:31 SimpleServer gracefully stopped

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
+SLEEP=0.5
+
 set -e
 echo "" > coverage.txt
 
+rm -f profile.out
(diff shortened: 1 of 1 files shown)


st0012:
@hachi8833 Thanks for notifying this, I haven't met this issue yet. However I recommend just run `go test ./... -race` because you don't need to generate test coverage report on your machine, which could make your test slower.

hachi8833:
Thank you for the checking!

PR #168, opened 2017-06-18
Add fsm comment

shes50103:
- Add comment on each state of FSM in lexer.go
- Rename initialize into initial

### DIFF
--- a/lexer/lexer.go
+++ b/lexer/lexer.go
@@ -22,9 +22,9 @@ func New(input string) *Lexer {
 	l.FSM = fsm.NewFSM(
 		"initial",
 		/*
-		Initial state is default state
-		Nosymbol state support we identify tok ':' is symbol or json
-		Method state support we identify tok 'class' is kind of keyword or Ident
+			Initial state is default state
+			Nosymbol state support we identify tok ':' is symbol or json
+			Method state support we identify tok 'class' is kind of keyword or Ident
 		*/
 		fsm.Events{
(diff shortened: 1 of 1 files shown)


shes50103:
I only add three lines comment to explain the purpose of states in FSM
Do I need to write more detail to  explain all the process in FSM?
Like how to enter each state and the effect after enter each state.

lexer/lexer.go
```diff
@@ -21,10 +21,15 @@ func New(input string) *Lexer {
 	l.readChar()
 	l.FSM = fsm.NewFSM(
 		"initial",
+		/*
+			Initial state is default state
+			Nosymbol state support we identify tok ':' is symbol or json
```
st0012: I think this sentence could be more expressive if we change it to:
`"Nosymbol state helps us identify tok ':' is for symbol or hash value"`
(`{ key:value}` this is not a json, it's a hash)

lexer/lexer.go
```diff
@@ -21,10 +21,15 @@ func New(input string) *Lexer {
...
+			Method state support we identify tok 'class' is kind of keyword or Ident
```
st0012: This should be changed to
`Method state helps us identify 'class' literal is a keyword or an identifier`

st0012:
@shes50103 You don't need to explain whole process, but you can provide related document links in the comment.

shes50103:
ok! I'll add the link of GitHub of FSM in comment

st0012:
@shes50103 Thanks for adding comments

PR #170, opened 2017-06-19
Add `$GOBY_ROOT` instruction in README

adlerhsieh:
I've moved instruction to [wiki](https://github.com/goby-lang/goby/wiki/Installation) since this section is getting longer. I also update the version number to align with the version on brew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,22 @@ It will have Ruby's syntax (I'll try to support all common syntaxes) but without
 
 ## Install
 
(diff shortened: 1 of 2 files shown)


st0012:
@adlerhsieh I think we should still put that section in readme, make user change pages is not a good experience. And I also think the section's length is in the acceptable range.

adlerhsieh:
Removed wiki page. Put `GOBY_ROOT` instructions in README.

Issue #177, opened 2017-06-26
Alpha REPL issues


hachi8833:
Checked on master branch, with the binary executable from `make install`.

* Empty echo backs are shown in some cases: 

<img width="150" alt="170627_0607_xarbp6" src="https://user-images.githubusercontent.com/4289625/27560657-f9cd3ed4-5aff-11e7-8fce-a8d851c7cce1.png">

showing `#=>` at the head of echo back would be better for usability.

* `exit` not working (needs ^C to exit instead for now):

```ruby
>> exit
UndefinedMethodError: Undefined Method 'exit' for <Instance of: Object>
ERROR: UndefinedMethodError: Undefined Method 'exit' for <Instance of: Object>
```

* Command history with up arrow key is not working for now:

```
>> ^[[A^[[A
```

* Pasting multiple lines works, but extra `>> >> >> >>` is shown when a multiple-line block is pasted:

![image](https://user-images.githubusercontent.com/4289625/27561066-9cee870c-5b01-11e7-83d6-897bd76b137d.png)

* Pasting statements at once returns multiple `>>` at the end; `>>` should be echo back by line instead:

![image](https://user-images.githubusercontent.com/4289625/27561268-733d16d4-5b02-11e7-9534-f45e25ccb4d9.png)

* Pasting [Sample-1](https://github.com/goby-lang/goby/blob/master/samples/sample-1.gb) got an error at line 20:

> puts(bar.get + baz.get + foo.get) #=> 16
UndefinedMethodError: Undefined Method '+' for
ERROR: UndefinedMethodError: Undefined Method '+' for

In addition, `puts(bar.get)` or `puts(bar)` after pasting returns nothing.

Looks like current REPL works on individual statements, but discards the generated object when multiple lines.

* Pasting [Sample-2](https://github.com/goby-lang/goby/blob/master/samples/sample-2.gb) returns nothing:

![image](https://user-images.githubusercontent.com/4289625/27562109-f07c4b76-5b05-11e7-9413-02012669a701.png)

In addition, I guess it reveals that several methods or expressions do not return any objects. 
(Sample-2 works on Ruby's irb)
Ruby's irb assumes that any expressions return objects.

st0012:
@hachi8833 Thanks for the report, and currently I can just say that assignment won't return anything. But other cases I need to investigate a little bit.

st0012:
@hachi8833 I just pushed a fix for this issue, pasting code should executes normally. I'm still working on fix formatting.

[...]

PR #178, opened 2017-06-28
Add test for igb.Start()

st0012:
Because I have problem simulating user input in tests, so I make
Start() to receive channel as argument and receives line from it.
This allow us to passing inputs by sending it through channel, currently
it works quite well.

### DIFF
--- a/goby.go
+++ b/goby.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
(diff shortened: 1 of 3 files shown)


st0012:
@hachi8833 It still behaves a little weird, but I can't spend too much time on it. So you can test it and update the remaining issue on #177

PR #183, opened 2017-07-01
Default arg value

st0012:
Three cases:

1. Single default arg: This performs same as Ruby
```ruby
def foo(x = 10)
  x
end

foo #=> 10
```

2. Two args and the first one has default value: You need to provide two args in this case, because the first one will overwrite `x` instead of being assigned to `y`. This performs different from Ruby
```ruby
def foo(x = 10, y)
  x + y
end

foo(100, 10) #=> 110
f00(10) #=> Argument Error
```

3. Two args but the last one has default value: The arg will be assigned to first one `x`, so `y`'s default value would still work.
```ruby
def foo(x, y = 10)
  x + y
end

foo(100) #=> 110
```

The reason it behaves like this is mostly because the bytecode can provide too few information for us to know what arg to assign, so I decide just assign them in defined order.


st0012:
@adlerhsieh  @hachi8833 Can you help checking the behavior and give me some feedback?

hachi8833:
The branch works as described and I think it is OK for simplification.

st0012:
@hachi8833 Can you help me document the behavior somewhere in the code after merging this PR?

Issue #182, opened 2017-07-01
[VM] Range API Design


Maxwell-Alexius:
According to [Ruby Ver. 2.4.1 Doc](https://ruby-doc.org/core-2.4.1/Range.html)

Range method implementation listed below:
```
range#first [Only lack commentary doc]
range#last [Only lack commentary doc]
range#each [Only lack commentary doc]
range#to_a [Only lack commentary doc]
range#step
range#include?
range#bsearch (binery search)
range#size
range#to_s
```

st0012:
LGTM, you can just start implementing them.

st0012:
@Maxwell-Alexius I'm closing this since the only unimplemented method already has its own issue.

PR #186, opened 2017-07-01
#182 Implementing Fundamental Range API

Maxwell-Alexius:
Currently struggling on implementing the `Range#bsearch` method, other methods are just fine.


st0012:
@Maxwell-Alexius Looks Can you also add some test cases for ranges that contain negative number? After that this should be ready to merge.

Maxwell-Alexius:
Yes, will do!

Maxwell-Alexius:
Tips: In Ruby 2.4.1, `range#step` do not accept negative step value, and it will only loop in ascending range.

Remark: If decide to implement negative step feature, feel free to discuss it or open another PR. For now, this PR implementing `range#step` do not accept negative step.

st0012:
@Maxwell-Alexius I see, current behavior is sufficient for most cases I guess. So is this ready to merge?

Maxwell-Alexius:
Yep, ready. (Almost forgot to remove WIP label off)

Issue #184, opened 2017-07-01
String API Refactor Error Test Case


Maxwell-Alexius:
According to [Ruby Ver. 2.4.1 Doc](https://ruby-doc.org/core-2.4.1/String.html)

String method implementation listed below:

Implemented and well-documented:
```
String#+
String#*
String#>
String#<
String#<=>
String#!=
String#capitalize
String#upcase
String#downcase
String#size
String#length
String#reverse
String#to_s
String#to_i
```

Haven't implemented yet:

- [ ] `String#<< (Same as #concat)`
- [x] `String#concat`
- [x] `String#[]`
- [x] `String#[]= (Replace the index of the string with another string or character)`
- [ ] `String#%`
- [ ] `String#index`
- [x] `String#to_a`
- [ ] `String#to_sym`
- [x] `String#count`
- [ ] `String#each_byte`
- [ ] `String#each_char`
- [ ] `String#each_line`
- [x] `String#empty?`
- [x] `String#eql?`
- [x] `String#start_with?`
- [x] `String#end_with?`
- [x] `String#insert`
- [x] `String#delete`
- [x] `String#chop (Returns a string which removes last character)`
- [ ] `String#chomp (Returns a new string which default removes separators like carriage returns if present in last of the string)`
- [ ] `String#dump (Returns raw string)`
- [ ] `String#center`
- [x] `String#ljust (String left justify)`
- [x] `String#rjust (String right justify)`
- [ ] `String#lstrip (Left Strip)`
- [ ] `String#rstrip (Right Strip)`
- [x] `String#replace`
- [x] `String#strip`
- [x] `String#slice`
- [x] `String#split`
- [ ] `String#squeeze (e.g. "Hello   I  like    Goby  Lang".squeeze => "Hello I like Goby Lang")`
- [ ] `String#succ`
- [ ] `String#upto`

Some methods from the doc I think is unnecessary or can be renamed?

- [ ] `String#chr (Returns the first character of the string, maybe rename to #start and then we can have the String#last method which calls the last character of the string)`
- [ ] `String#clear (Returns empty string)`

Not sure to implement for now:

- [ ] `String#to_f`
- [ ] `String#to_c`
- [ ] `String#=~`
- [x] `String#gsub`
- [ ] `String#encode`
- [ ] `String#valid_encoding?`

st0012:
@Maxwell-Alexius Can you use `- [ ]` to make list checkable? Then I can just check what I think is needed.

[...]

PR #189, opened 2017-07-02
#184 Implementing Fundamental String API

Maxwell-Alexius:
This PR aims to extend the string API method designed in #184

### DIFF
--- a/vm/string.go
+++ b/vm/string.go
@@ -31,7 +31,7 @@ type StringObject struct {
 }
 
 func (vm *VM) initStringObject(value string) *StringObject {
-	replacer := strings.NewReplacer("\\n", "\n", "\\r", "\r", "\\t", "\t", "\\v", "\v", "\\\\", "\\")
+	replacer := strings.NewReplacer("\\n", "\n", "\\r", "\r", "\\t", "\t", "\\v", "\v", "\\f", "\f", "\\\\", "\\")
 	return &StringObject{
 		baseObj: &baseObj{class: vm.topLevelClass(stringClass)},
 		Value:   replacer.Replace(value),
@@ -598,9 +598,8 @@ func builtInStringClassMethods() []*BuiltInMethodObject {
 						if -indexValue > strLength {
 							return initErrorObject(ArgumentErrorClass, "Index value out of range. got=%v", strconv.Itoa(indexValue))
 						}
-
-						result := str[:strLength+indexValue] + replaceStrValue + str[strLength+indexValue+1:]
-						return t.vm.initStringObject(result)
+						// Change to positive index to replace the string
+						indexValue += strLength
 					}
 
 					if strLength == indexValue {
@@ -810,7 +809,8 @@ func builtInStringClassMethods() []*BuiltInMethodObject {
 						} else if -indexValue == strLength+1 {
 							return t.vm.initStringObject(insertStr.Value + str)
 						}
-						return t.vm.initStringObject(str[:strLength+indexValue] + insertStr.Value + str[strLength+indexValue:])
+						// Change it to positive index value to replace the string via index
+						indexValue += strLength
 					}
 
 					if strLength < indexValue {
@@ -1016,7 +1016,7 @@ func builtInStringClassMethods() []*BuiltInMethodObject {
 			// ```ruby
 			// "Hello World".split("o") # => ["Hell", " W", "rld"]
 			// "Goby".split("")         # => ["G", "o", "b", "y"]
-			// # TODO: Whitespace carriage return case
+			// "Hello\nWorld\nGoby".split("o") # => ["Hello", "World", "Goby"]
 			// ```
 			//
 			// @return [Array]
@@ -1130,7 +1130,7 @@ func builtInStringClassMethods() []*BuiltInMethodObject {
 						return t.vm.initStringObject(string([]rune(str)[intValue]))
 
 					default:
-						return initErrorObject(ArgumentErrorClass, "Expect slice range is Range or Integer type. got=%T", args[0])
+						return initErrorObject(ArgumentErrorClass, "Expect slice range to be Range or Integer. got=%T", args[0])
 					}
 				}
 			},
--- a/vm/string.go
+++ b/vm/string.go
@@ -61,10 +61,11 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
 
 					leftValue := receiver.(*StringObject).Value
-					right, ok := args[0].(*StringObject)
+					r := args[0]
+					right, ok := r.(*StringObject)
 
 					if !ok {
-						return wrongTypeError(receiver.Class())
+						return initErrorObject(TypeErrorClass, "Expect argument to be String. got=%T", r)
 					}
 
 					rightValue := right.Value
@@ -119,10 +120,11 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
 
 					leftValue := receiver.(*StringObject).Value
-					right, ok := args[0].(*StringObject)
+					r := args[0]
+					right, ok := r.(*StringObject)
 
 					if !ok {
-						return wrongTypeError(receiver.Class())
+						return initErrorObject(TypeErrorClass, "Expect argument to be String. got=%T", r)
 					}
 
 					rightValue := right.Value
@@ -148,10 +150,11 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
 
 					leftValue := receiver.(*StringObject).Value
-					right, ok := args[0].(*StringObject)
+					r := args[0]
+					right, ok := r.(*StringObject)
 
 					if !ok {
-						return wrongTypeError(receiver.Class())
+						return initErrorObject(TypeErrorClass, "Expect argument to be String. got=%T", r)
 					}
 
 					rightValue := right.Value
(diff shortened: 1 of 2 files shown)


Maxwell-Alexius:
@st0012 Can you help me check out Travis CI Error log?

I tried to implement the similar kind of error test suit from `integer_test.go` to `string_test.go`. `integer_test.go` seems to work fine.

However, the `string_test.go` can't pass because the error object pointer class seems to pointing at `nil` rather than `string` class (See travis CI error log):

```
&{<nil> expect argument to be String type}
```

vm/string.go
```diff
@@ -55,6 +56,16 @@ func (s *StringObject) returnClass() Class {
 	return s.Class
 }
 
+func (s *StringObject) toArray() *ArrayObject {
+	elems := []Object{}
+
+	for i := 0; i < len(s.Value); i++ {
+		elems = append(elems, initIntegerObject(i))
```
st0012: Why would the array hold a list of integers instead of single character?

vm/string_test.go
```diff
@@ -59,18 +73,92 @@ func TestEvalInfixStringExpression(t *testing.T) {
...
+		{`"1234567890".slice(11..1)`, nil},
+		{`"1234567890".slice(11..-1)`, nil},
+		{`"1234567890".slice(-10..1)`, "12"},
+		{`"1234567890".slice(-5..1)`, ""},
+		{`"1234567890".slice(-10..-1)`, "1234567890"},
+		{`"1234567890".slice(-10..-11)`, ""},
+		{`"1234567890".slice(1..-1)`, "234567890"},
+		{`"1234567890".slice(1..-1234)`, ""},
+		{`"1234567890".slice(-11..5)`, nil},
+		{`"1234567890".slice(-11..-12)`, nil},
```
st0012: @Maxwell-Alexius  The error happens in this test case, it seems like it'll still returns string object instead of `Null`

st0012:
@Maxwell-Alexius BTW can you group string's infix test cases into different test functions like `TestEvalStringComparison`..etc. by their test subject? Current test case list is too long for debugging.

Maxwell-Alexius:
@st0012 I shall open another PR to rearrange those test

vm/string_test.go
Maxwell-Alexius: Finally, after correcting this test case, this problem also solved other error test case implementation. Many thanks =) Then I shall start filling other test case to boost up the test coverage

st0012: 👍 looking forward to merge this

Maxwell-Alexius:
Mention one more thing: In this PR, the test for `String#[]` and `String#[]=` cannot be fully implemented for the reason stated in #222

vm/string_test.go
```diff
@@ -29,6 +29,7 @@ func TestStringConversion(t *testing.T) {
 		{`"string".to_i`, 0},
 		{`"123string123".to_i`, 123},
 		{`"string123".to_i`, 0},
+		{`"Goby".to_a`, initArrayObject([]Object{initStringObject("G"), initStringObject("o"), initStringObject("b"), initStringObject("y")})},
```
Maxwell-Alexius: @st0012 Other string test cases works fine. Except, in the `String#to_a` test, when I tried to change the first element of expected value in `initArrayObject`, which is `initStringObject("G")` to other kind of value (e.g. `initStringObject("H")`), it didn't raise any error but passed the test

Maxwell-Alexius:
Sorry, I accidentally deleted the branch in my `origin`

vm/string.go
```diff
@@ -471,6 +472,719 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
...
+						return initErrorObject(TypeErrorClass, "Expect Integer. got=%T (%+v)", args[0], args[0])
+					}
+
+					indexValue := index.Value
+
+					if indexValue < 0 {
+						if -indexValue > len(str) {
+							return NULL
+						}
+						return t.vm.initStringObject(string([]rune(str)[len(str)+indexValue]))
```
st0012: Why we need the string->rune->string transition? What's the difference between
```go
string([]rune(str)[len(str)+indexValue])
```
and
```go
string[len(str)+indexValue]
```

vm/string.go
```diff
@@ -471,6 +472,719 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
...
+					if len(args) != 1 {
+						return initErrorObject(ArgumentErrorClass, "Expect 1 argument. got=%v", strconv.Itoa(len(args)))
+					}
+
+					str := receiver.(*StringObject).Value
+					i := args[0]
+					index, ok := i.(*IntegerObject)
+
+					if !ok {
+						return initErrorObject(TypeErrorClass, "Expect Integer. got=%T (%+v)", args[0], args[0])
```
st0012: Just pass `i` instead of `args[0]`?

st0012: And why is this's index type error message different from `[]=`? I think they can share same index error message?

vm/string.go
```diff
@@ -471,6 +472,719 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
...
+
+					if strLength < indexValue {
+						return initErrorObject(ArgumentErrorClass, "Index value out of range. got=%v", strconv.Itoa(indexValue))
+					}
+
+					r := args[1]
+					replaceStr, ok := r.(*StringObject)
+
+					if !ok {
+						return initErrorObject(TypeErrorClass, "Expect to assign String type value. got=%T", r)
```
st0012: `Expect argument to be String`

vm/string.go
```diff
@@ -471,6 +472,719 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
...
+					}
+					replaceStrValue := replaceStr.Value
+
+					// Negative Index Case
+					if indexValue < 0 {
+						if -indexValue > strLength {
+							return initErrorObject(ArgumentErrorClass, "Index value out of range. got=%v", strconv.Itoa(indexValue))
+						}
+
+						result := str[:strLength+indexValue] + replaceStrValue + str[strLength+indexValue+1:]
```
st0012: Maybe add a comment here for more detailed explanation?

vm/string.go
```diff
@@ -471,6 +472,719 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
...
+			// @return [Boolean]
+			Name: "end_with",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+
+					str := receiver.(*StringObject).Value
+					compareStr, ok := args[0].(*StringObject)
+
+					if !ok {
+						return wrongTypeError(receiver.Class())
```
st0012: Why don't you update error report function here?

vm/string.go
hachi8833: I'm afraid implementing rune type is a kind of future work, because this is a comprehensive issue on Goby and we need survey before doing that.

vm/string.go
Maxwell-Alexius: Reference:
https://stackoverflow.com/questions/15018545/how-to-index-characters-in-a-golang-string

This case if for `ASCII` only:
```
string[len(str)+indexValue]
```

However, the transition to `rune` type can be used to `UTF-8` condition, try the example below:
```
package main

import "fmt"

func main() {
    fmt.Println(string("Hello, 世界"[1]))         // ASCII only
    fmt.Println(string("Hello, 世界"[8]))         // ASCII only
    fmt.Println(string([]rune("Hello, 世界")[1])) // UTF-8
    fmt.Println(string([]rune("Hello, 世界")[8])) // UTF-8
}
```

Got:
```
e
¸
e
界
```

vm/string.go
Maxwell-Alexius: Ok, I'll do that

hachi8833:
I realized my PR #235 changed the structure of string.go. Sorry.

Maxwell-Alexius:
@hachi8833  It's okay, I already rebased my branch including your PR. But first, I need to push the updated version for you guys to help me review it. Thanks a lot!

vm/string_test.go
```diff
@@ -76,6 +131,61 @@ func TestStringOperation(t *testing.T) {
...
+	}
+}
+
+func TestStringOperationFail(t *testing.T) {
+	vm := initTestVM()
+	testsFail := []struct {
+		input    string
+		expected *Error
+	}{
+		{`"Taipei" + 101`, initErrorObject(TypeErrorClass, "Expect argument to be String. got=%T", vm.initIntegerObject(101))},
```
st0012: Here you should write down the whole error messages. Because if there's an issue in `initErrorObject` you'll never find out in these tests. So the `expected` field's type should be string`.

vm/string.go
```diff
@@ -117,10 +133,11 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
 
 					leftValue := receiver.(*StringObject).Value
-					right, ok := args[0].(*StringObject)
+					r := args[0]
+					right, ok := r.(*StringObject)
 
 					if !ok {
-						return wrongTypeError(receiver.Class())
+						return initErrorObject(TypeErrorClass, "Expect argument to be String. got=%T", r)
```
st0012: You shouldn't use `%T` because that prints object's type in Go, which might be `*IntegerObject` but what we want is `Integer`. So just use `%s` and `r.Class().Name`.

st0012: And if you test this by matching the raw string, you would've found this issue.

vm/string_test.go
st0012: I think you just followed wrong examples, when it comes to testing error, you should take a look at `vm/error_test.go` first. Sorry didn't notify you earlier.

st0012:
@Maxwell-Alexius I'll merge this first because these methods are already functional. And you should open another PR for changing error testing.

Issue #191, opened 2017-07-02
Array#at causes panic when passing `-1`


hachi8833:
```ruby
a = [1, 2, 3, 4, 5]
a.at(-1)  # => panic
a[-1]      # => panic
```

Maxwell-Alexius:
I may open a PR to implement negative index condition for `Array#at`, `Array#[]` and `Array#[]=`

hachi8833:
Confirmed your fix works fine! I close the issue.

PR #193, opened 2017-07-02
#191 [VM Part] Add Negative Index Support for Array API Method

Maxwell-Alexius:
Negative index support for `Array#[]`, `Array#[]=` and `Array#at`


vm/array_test.go
```diff
@@ -155,6 +171,15 @@ func TestEvalArrayIndex(t *testing.T) {
 			[1, "a", 10, 5].at(1)
 		`, "a"},
 		{`
+			[1, "a", 10, 5].at(4)
+		`, NULL},
```
st0012: Shouldn't this also be `nil` too? Like above tests.

st0012:
@Maxwell-Alexius GJ!

PR #195, opened 2017-07-02
Add and update API documents

hachi8833:
Mainly added doc to  Hash, and updated some others.

### DIFF
--- a/vm/hash.go
+++ b/vm/hash.go
@@ -41,6 +41,7 @@ type RHash struct {
 // **Note:**
 // - The order of key-value pairs are **not** preserved.
 // - Operator `=>` is not supported.
+// - `Hash.new` is not supported.
 type HashObject struct {
 	Class *RHash
 	Pairs map[string]Object
--- a/vm/hash.go
+++ b/vm/hash.go
@@ -15,7 +15,32 @@ type RHash struct {
 	*BaseClass
 }
 
-// HashObject represents hash instances
+// Hash is a collection of key-value pair, which works like a dictionary.
+// Hash literal is represented with curly brackets `{ }` like `{ key: value }`.
+// Each key of the hash is unique and cannot be duplicate within the hash.
+// Adding a leading space and a trailing space within curly brackets are preferable.
+//
+// - **Key:** an alphanumeric word that starts with alphabet, without containing space and punctuations.
+// Underscore `_` can also be used within the key.
+// String literal like "mickey mouse" cannot be used as a hash key.
+// The internal key is actually a String and **not a Symbol** for now (TBD).
+// Thus only a String object or a string literal should be used when referencing with `[ ]`.
+//
+// ```ruby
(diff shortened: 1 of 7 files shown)


vm/hash.go
```diff
@@ -127,6 +167,18 @@ var builtinHashInstanceMethods = []*BuiltInMethodObject{
...
+		//
+		// ```Ruby
+		// h = { a: 1, b: "2", c: [1, 2, 3], d: { k: 'v' } }
+		// h['a'] = 1          #=> 1
+		// h['b'] = "2"        #=> "2"
+		// h['c'] = [1, 2, 3]  #=> [1, 2, 3]
+		// h['d'] = { k: 'v' } #=> { k: 'v' }
+		// ```
+		//
+		// @return [value]
```
adlerhsieh: The name in the brackets is the class of the returned value, so it should be `[Object]` instead. The `value` can go to explanation afterwards.

vm/hash.go
```diff
@@ -152,6 +204,14 @@ var builtinHashInstanceMethods = []*BuiltInMethodObject{
 		},
 	},
 	{
+		// Return the number of key-value pairs of the hash.
```
adlerhsieh: Probably make it "Returns"?

vm/range.go
```diff
@@ -67,6 +101,19 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+		// ```ruby
+		// (1..5).to_a     # => [1, 2, 3, 4, 5]
+		// (1..5).to_a[2]  # => 3
+		// ```
+		//
+		// **Note:**
+		// - Only `do`-`end` block is supported for now: `{ }` block is unavailable.
+		// - Three-dot range `...` is not supported yet.
+		//
+		// @return [nil]
```
adlerhsieh: Should be capitalized class name `Null`.

vm/string.go
```diff
@@ -17,6 +17,22 @@ type RString struct {
...
+// ```ruby
+// a = "Three"
+// b = 'zero'
+// c = '漢'
+// d = 'Tiếng Việt'
+// e = "😏️️"
+// ```
+//
+// **Note:**
+// Currently, manipulations are based upon Golang's Unicode manipulations.
```
adlerhsieh: Add a blank line between the sentence and the bullet points.

vm/hash.go
hachi8833: Fixed

vm/hash.go
hachi8833: Fixed

vm/range.go
hachi8833: Fixed

vm/string.go
hachi8833: Fixed

st0012:
@hachi8833 Nice work!

PR #197, opened 2017-07-02
Add `String#include`

hachi8833:
My first method implementation.


st0012:
@hachi8833 Happy to see your first method contribution, it looks good to me, well done 😎

Issue #196, opened 2017-07-02
Special characters such as `\n` are not interpreted


hachi8833:
Special characters with backslash such as `\n` `\r` `\t` `\v` are not treated. Looks like string literals are actually just a byte stream.

* Goby

```ruby
a = "Hello\nWorld"
puts(a)  
# => "Hello\nWorld"
```

* Ruby

```ruby
a = "Hello\nWorld"
puts(a)  
# => Hello
# => World
```

Maxwell-Alexius:
#189 meets some sort of same problem, such as when implementing the `String#strip`, it need to handle the case such that:

```ruby
"\n\t Hello World\t \r\n".strip # => "Hello World"
```

hachi8833:
Actually, I met the same issue when implementing `String#strip` ;-)

Well, I suppose we should also change `string` type in Golang to `rune` type.

st0012:
I'll let @Maxwell-Alexius work on this first.

Maxwell-Alexius:
@st0012 @hachi8833 I'm not sure whether to also convert `string` type to `rune` type in `initStringObject`? What are your opinions?

st0012:
I'm not very familiar with `rune` type, need to spend some to do some research.

Maxwell-Alexius:
Me, too. There's a lot to learn, then leave this issue for a moment. (I'll keep on implementing other string api in #184 )

st0012:
@hachi8833 I think @Maxwell-Alexius already fixed this, can you help verify?

hachi8833:
Confirmed `\n` `\t` works now!
I close the issue.

PR #199, opened 2017-07-03
#196 Init string handled with carriage returns or whitespace case

Maxwell-Alexius:
@hachi8833 @st0012  This PR aims to let special chars including `\r, \t, \n, \v, \\` being interpreted


st0012:
@Maxwell-Alexius I think there should be a better solution instead of replacing them one by one, I am also working on this.

st0012:
@Maxwell-Alexius We should use `Replacer` instead of replace them one by one. It'll be more readable and performant. [Reference](https://www.dotnetperls.com/replace-go)

PR #201, opened 2017-07-03
Use checkExpected for all test result checking.

st0012:
This should simplify the way we check test result and make tests more unified.


st0012:
@Maxwell-Alexius You might want to check this, since you'll write a lot tests when implementing APIs

PR #204, opened 2017-07-03
Returns error when method call's argument count not matching.

### LINKED ISSUE #203: Goby not returning error when calling a method without enough arguments
Current behavior:

```ruby
def foo(x); end
foo #=> returns nil
```

But it should raise an error like: `"ArgumentError: Expect 1 args for method 'foo'. got: 0"`
And there's a big obstacle for implementing this: default argument value

```ruby
def foo(x = 10, y); end
foo(10) #=> Should raise error

def bar(x, y = 10); end
bar(10) #=> Shouldn't raise error
```

cc @hachi8833

st0012:
This fixes #203

### DIFF
--- a/vm/error_test.go
+++ b/vm/error_test.go
@@ -2,20 +2,6 @@ package vm
 
 import "testing"
 
-func TestArgumentError(t *testing.T) {
-	vm := initTestVM()
-	evaluated := vm.testEval(t, "[].count(5,4,3)")
-	err, ok := evaluated.(*Error)
-	if !ok {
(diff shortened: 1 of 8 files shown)


vm/error_test.go
```diff
@@ -112,3 +98,56 @@ func TestUnsupportedMethodError(t *testing.T) {
 		}
 	}
 }
+
+func TestArgumentError(t *testing.T) {
```
st0012: @Maxwell-Alexius  @hachi8833 Can you help me check if these test cases are enough?

PR #207, opened 2017-07-04
Add compiler package

st0012:
## Purpose: 

- Use compiler package to hide compilation's implementation detail.
- Use compiler to export compiled instruction sets without expose code generator.


st0012:
The test coverage dropped significantly because those bytecode string related methods won't be tested now.

Issue #202, opened 2017-07-03
Range#bsearch Implementation


Maxwell-Alexius:
This issue aims to implementing `Range#bsearch` in Goby Lang

st0012:
This is finished, thanks @Maxwell-Alexius!

PR #209, opened 2017-07-04
#202 Implementing Range#bsearch API

Maxwell-Alexius:
This PR aims to implement binary search method in Range API from Ruby 2.4.1 to Goby Lang

### DIFF
--- a/vm/range.go
+++ b/vm/range.go
@@ -1,8 +1,6 @@
 package vm
 
-import (
-	"fmt"
-)
+import "fmt"
 
 var (
 	rangeClass *RClass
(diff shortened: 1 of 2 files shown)


vm/range.go
```diff
@@ -306,4 +306,109 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+							return NULL
+						}
+
+						if value > 0 {
+							start = mid + 1
+						} else {
+							end = mid - 1
+						}
+					default:
+						return newError("Must be integer or boolean type. got=%T", result)
```
st0012: Please use this method instead https://github.com/goby-lang/goby/blob/master/vm/error.go#L59

vm/range.go
```diff
@@ -306,4 +306,109 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+							return initIntegerObject(pivot)
+						}
+
+						if value {
+							end = mid - 1
+						} else {
+							start = mid + 1
+						}
+					case *IntegerObject:
+						fmt.Println(start + end)
```
st0012: Remember to remove this line

vm/range.go
```diff
@@ -306,4 +306,109 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+				start := ran.Start
+				end := ran.End
+				var mid int
+				pivot := -1
+
+				for {
+					if (start+end)%2 != 0 {
+						mid = (start+end)/2 + 1
+					} else {
+						mid = (start + end) / 2
```
st0012: Should there be spaces around `+`? Make it consistent like above lines.

vm/range.go
```diff
@@ -306,4 +306,109 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+					}
+					result := t.builtInMethodYield(blockFrame, initIntegerObject(mid))
+
+					switch result.Target.(type) {
```
st0012: You can write `r := result.Target.(type)` then `r` will be assert to the matched type.

vm/range.go
```diff
@@ -306,4 +306,109 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+					case *BooleanObject:
+						value := result.Target.(*BooleanObject).Value
```
st0012: So here you can just `value := r.Value`, Go already assumes `r` is `*BooleanObject` type.

vm/range.go
Maxwell-Alexius: I found out that when executing:

```
$ go test ./...
```

It just auto-corrected the second `mid` assignment which makes the space wraps around the `+` sign

vm/range.go
Maxwell-Alexius: Can I extend something like `WrongTypeErrorClass`?

Maxwell-Alexius: Oops, I saw `TypeError` implemented, sorry XD

vm/range.go
st0012: You need to type -s -w

vm/range.go
```diff
@@ -306,4 +304,106 @@ var builtInRangeInstanceMethods = []*BuiltInMethodObject{
...
+							return NULL
+						}
+
+						if r.Value > 0 {
+							start = mid + 1
+						} else {
+							end = mid - 1
+						}
+					default:
+						return initErrorObject(TypeErrorClass, "Expect Integer or Boolean type. got=%v", r)
```
Maxwell-Alexius: @st0012 Can you help me check it. I'm not quite sure this error implementation is right or wrong.

st0012: Yes, you can open REPL and make this method fail, if the REPL doesn't crash and you see the error message means it works.

Maxwell-Alexius: Thanks, let me try it

st0012:
LGTM, @hachi8833 do you want to review this?

PR #211, opened 2017-07-04
#184 Rearrange String Test

Maxwell-Alexius:


vm/string_test.go
```diff
@@ -39,34 +34,156 @@ func TestEvalInfixStringExpression(t *testing.T) {
...
+		{`"Minus" * 1`, "Minus"},
+	}
+
+	for _, tt := range tests {
+		evaluated := testEval(t, tt.input)
+		checkExpected(t, evaluated, tt.expected)
+	}
+}
+
+func TestCapitalizeString(t *testing.T) {
```
st0012: For naming consistency this should be `TestStringCapitalization` or `TestCapitalizingString`. I think the later one is more easy to write and understand. Same rule applies to other functions.

vm/string_test.go
```diff
@@ -39,34 +34,156 @@ func TestEvalInfixStringExpression(t *testing.T) {
...
 		{`"string123".to_i`, 0},
+	}
+
+	for _, tt := range tests {
+		evaluated := testEval(t, tt.input)
+		checkExpected(t, evaluated, tt.expected)
+	}
+}
+
+func TestEvalInfixStringExpression(t *testing.T) {
```
st0012: This should be renamed to `TestChainingStringMethods` I think

Issue #213, opened 2017-07-04
`include ModuleName` on top level causes panic


hachi8833:
* calling `include` on top level causes panic:

```ruby
module MyModule
  class MyClass
    def test(a)
      puts(a)
    end
  end
end

include MyModule #=> panic
```

hachi8833:
Adding `""` or `''` to module name on `include` also causes panic:

```ruby
module MyModule
  class MyClass
    def test(a)
      puts(a)
    end
  end
end

class MyClass2
  include 'MyModule' #=> panic
end
```

st0012:
@hachi8833 The first one is bug, but the second one is not what I want to support.

hachi8833:
Yes, just showing error is sufficient in the second case.

st0012:
@hachi8833 Can you open a new issue for those panics that should be replaced by normal error reporting? Make it a checklist and with the link of the github source code page (with line number). Then we can close issue once first case is confirmed fixed.

st0012:
I saw the new issue and I'm closing this now

PR #214, opened 2017-07-04
Fix vm issues

st0012:
I spotted some issues inside VM so I just fix them all in here.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -367,7 +367,7 @@ var builtinCommonInstanceMethods = []*BuiltInMethodObject{
 				file, err := ioutil.ReadFile(filepath + ".gb")
 
 				if err != nil {
-					return initErrorObject(InternalErrorClass,  err.Error())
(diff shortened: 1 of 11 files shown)


vm/object.go
```diff
@@ -7,6 +7,10 @@ import (
 var mainObj *RObject
 
 func init() {
```
st0012: We shouldn't avoid use this `init` function and package level variables, I'll fix these issues in future commits.

vm/class.go
```diff
@@ -708,8 +707,18 @@ var builtinClassClassMethods = []*BuiltInMethodObject{
...
 			return func(t *thread, args []Object, blockFrame *callFrame) Object {
+				var class *RClass
 				module := args[0].(*RClass)
-				class := receiver.(*RClass)
+
+				switch r := receiver.(type) {
+				case *RClass:
+					class = r
+				case *RObject:
+					if r.Class == objectClass {
```
st0012: @hachi8833 Here should fix #213

PR #216, opened 2017-07-05
WIP: Regexp class 

hachi8833:
Just a trial for now. 

* Using https://github.com/dlclark/regexp2/ : almost all equivalent to Onigmo, multilingual-conscious
* Implemented as far as I can, mimicrying the recent implementation around compiler
* Current method `compile` is still a dummy
* `make test` works safely
* But regexp literal like `a = /aaa.*/` still does not work: this blocks me so far
* Unsure how to perform `godep save`
* Tests have not been added yet

(The branch I pushed just before was obsolete. Sorry)


vm/instruction.go
```diff
@@ -231,6 +239,13 @@ var builtInActions = map[operationType]*action{
 			t.stack.push(&Pointer{cf.self})
 		},
 	},
+	bytecode.PutRegexp: {
```
st0012: Why can't regex uses `putobject` instruction?

hachi8833: I'll look into this

Issue #218, opened 2017-07-05
Refactor VM for package level variable usage.


st0012:
Currently `VM` uses `init()` method and package level variables for built in classes. This is a bad way to write Go program (my fault), because it'll make code harder to trace and may cause unintentional mutation on those variables and breaks the whole program.

For those who want to know more, [this article](http://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html) has clear explanation.

cc @goby-lang/maintainers

PR #219, opened 2017-07-05
Refactor VM as #218 describes

st0012:
This is the refactoring #218 described

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -85,9 +85,8 @@ func (vm *VM) initArrayObject(elements []Object) *ArrayObject {
 	return &ArrayObject{Elements: elements, Class: vm.builtInClasses["Array"]}
 }
 
-func initArrayClass() *RClass {
-	bc := &BaseClass{Name: "Array", ClassMethods: newEnvironment(), Methods: newEnvironment(), Class: classClass, pseudoSuperClass: objectClass, superClass: objectClass}
-	ac := &RClass{BaseClass: bc}
+func (vm *VM) initArrayClass() *RClass {
+	ac := vm.initializeClass("Array", false)
 	ac.setBuiltInMethods(builtinArrayInstanceMethods, false)
 	ac.setBuiltInMethods(builtInArrayClassMethods(), true)
 	return ac
--- a/vm/array.go
+++ b/vm/array.go
@@ -224,7 +224,7 @@ var builtinArrayInstanceMethods = []*BuiltInMethodObject{
 				}
 
 				arr := receiver.(*ArrayObject)
-				return initIntegerObject(arr.length())
+				return t.vm.initIntegerObject(arr.length())
 			}
 		},
 	},
@@ -325,7 +325,7 @@ var builtinArrayInstanceMethods = []*BuiltInMethodObject{
 				}
 
 				for i := range arr.Elements {
-					t.builtInMethodYield(blockFrame, initIntegerObject(i))
+					t.builtInMethodYield(blockFrame, t.vm.initIntegerObject(i))
 				}
 				return arr
 			}
@@ -509,11 +509,11 @@ var builtinArrayInstanceMethods = []*BuiltInMethodObject{
 						}
 					}
 
-					return initIntegerObject(count)
+					return t.vm.initIntegerObject(count)
 				}
 
 				if len(args) == 0 {
-					return initIntegerObject(len(arr.Elements))
+					return t.vm.initIntegerObject(len(arr.Elements))
 				}
 
 				arg := args[0]
@@ -542,7 +542,7 @@ var builtinArrayInstanceMethods = []*BuiltInMethodObject{
 					}
 				}
 
-				return initIntegerObject(count)
+				return t.vm.initIntegerObject(count)
 			}
 		},
 	},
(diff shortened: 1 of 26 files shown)


st0012:
@goby-lang/maintainers so this means I should continue this work?

Maxwell-Alexius:
Yes, I think that's a good approach

vm/range_test.go
```diff
@@ -240,37 +240,37 @@ func TestInclude(t *testing.T) {
...
-		`, TRUE},
+		`, true},
 		{`
 		(5..10).include(5)
-		`, TRUE},
+		`, true},
 		{`
 		(5..10).include(4)
-		`, FALSE},
+		`, false},
```
st0012: @Maxwell-Alexius Use go's primitive type for testing, we don't need to init goby objects when running tests.

Maxwell-Alexius: Got it

vm/array_test.go
```diff
@@ -311,115 +319,120 @@ func TestEachIndexMethod(t *testing.T) {
...
 
 	for i, tt := range tests {
-		evaluated := testEval(t, tt.input)
+		v := initTestVM()
+		evaluated := v.testEval(t, tt.input)
 		checkExpected(t, i, evaluated, tt.expected)
 	}
 }
 
 func TestMapMethod(t *testing.T) {
```
st0012: @goby-lang/maintainers Array tests this way

vm/array.go
```diff
@@ -90,533 +81,547 @@ func (a *ArrayObject) shift() Object {
 	return value
 }
 
-// initArrayObject returns an array that contains given objects
-func initArrayObject(elements []Object) *ArrayObject {
-	return &ArrayObject{Elements: elements, Class: arrayClass}
+func (vm *VM) initArrayObject(elements []Object) *ArrayObject {
```
st0012: Now we need vm to initialize built in type's objects, because their classes are stored in vm.

vm/boolean.go
```diff
@@ -44,144 +37,160 @@ func (b *BooleanObject) equal(e *BooleanObject) bool {
...
+			// @return [Boolean]
+			Name: "&&",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+
+					leftValue := receiver.(*BooleanObject).Value
+					right, ok := args[0].(*BooleanObject)
+
+					if !ok {
+						return wrongTypeError(receiver.returnClass())
```
st0012: I just found that we can pass receiver's class directly instead of specifying the class. My bad.

vm/vm_test.go
```diff
@@ -74,6 +74,22 @@ foo
 	}
 }
 
+func initTestVM() *VM {
```
st0012: I separate vm's initialization from `testEval` because in this way we can access the vm's internal state if we want to.

vm/vm.go
```diff
@@ -188,36 +191,48 @@ func (vm *VM) GetREPLResult() string {
...
+		vm.initRangeClass(),
+		vm.initMethodClass(),
+		vm.initializeChannelClass(),
 	}
 
+	vm.initErrorClasses()
+
 	for _, c := range builtInClasses {
 		p := &Pointer{Target: c}
 		constants[c.ReturnName()] = p
```
st0012: We currently store built in classes in both vm's `constants` and `builtInClasses`, I'll fix this in #224

vm/class.go
```diff
@@ -7,13 +7,22 @@ import (
 	"time"
 )
 
-var (
-	objectClass *RClass
-	classClass  *RClass
+const (
+	objectClass  = "Object"
```
st0012: I use constant to store built in classes' names.

st0012:
@goby-lang/maintainers I know this is a huge PR and you can't possibly review it, but I have commented some significant changes or decision, please do read them and give me some feedback 🙇

vm/array.go
```diff
@@ -90,533 +81,547 @@ func (a *ArrayObject) shift() Object {
...
-		// a[-1] # => "c"
-		// a[-3] # => "a"
-		// a[-7] # => nil
-		// ```
-		Name: "[]",
-		Fn: func(receiver Object) builtinMethodBody {
-			return func(t *thread, args []Object, blockFrame *callFrame) Object {
-				if len(args) != 1 {
-					return &Error{Message: "Expect 1 arguments. got=%d" + string(len(args))}
+func builtInArrayClassMethods() []*BuiltInMethodObject {
```
st0012: @adlerhsieh Now all methods are stored in functions (both built in classes and standard libs), so you might need to change the API document parser to adjust this.

vm/channel.go
```diff
@@ -39,28 +31,32 @@ func (m *objectMap) storeObj(obj Object) int {
...
 // ChannelObject represents a goby channel, which carries a golang channel
 type ChannelObject struct {
 	id    int
 	Class *RClass
 	Chan  chan int
 }
 
-func initializeChannelClass() {
-	class := initializeClass("Channel", false)
+func (vm *VM) initializeChannelClass() *RClass {
```
hachi8833: `initChannelClass()` might be better for consistency

st0012: Thanks! pushed

PR #223, opened 2017-07-06
#202 Re-implementation of Range#bsearch With Test Case

Maxwell-Alexius:
Fix some test bug and add error testing with `Range#bsearch`

### DIFF
--- a/vm/range_test.go
+++ b/vm/range_test.go
@@ -441,12 +441,12 @@ func TestBsearchFail(t *testing.T) {
 		input    string
 		expected *Error
 	}{
-		//{`
-		//ary = [0, 4, 7, 10, 12]
-		//(0..4).bsearch do |i|
-		//	"Binary Search"
-		//end
-		//`, initErrorObject(TypeErrorClass, "Expect Integer or Boolean type. got=%T", initStringObject("Binary Search"))},
+		{`
+		ary = [0, 4, 7, 10, 12]
+		(0..4).bsearch do |i|
+			"Binary Search"
+		end
+		`, initErrorObject(TypeErrorClass, "Expect Integer or Boolean type. got=%T", vm.initStringObject("Binary Search"))},
 	}
 
 	for _, tt := range testsFail {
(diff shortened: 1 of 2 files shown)


Maxwell-Alexius:
`Range#bsearch` tests are ensured that `checkExpected` method works fine

st0012:
@Maxwell-Alexius Can you rebase this and apply the newest test approach like [here](https://github.com/goby-lang/goby/blob/master/vm/array_test.go#L51)?

vm/range_test.go
```diff
@@ -294,3 +294,169 @@ func TestInclude(t *testing.T) {
...
+		input    string
+		expected *Error
+	}{
+		{`
+		ary = [0, 4, 7, 10, 12]
+		(0..4).bsearch do |i|
+			"Binary Search"
+		end
+		`, initErrorObject(TypeErrorClass, "Expect Integer or Boolean type. got=%T", vm.initStringObject("Binary Search"))},
+	}
```
Maxwell-Alexius: @st0012 Reformatted, however, I need to put the `vm := initTestVM()` expression in the beginning of the function testing because I need to use it for error testing

st0012: That's ok

Maxwell-Alexius:
CFP checked, only two case to pop out the call frame:

- `ran.Start` is larger than `ran.End`
- the result of call frame is not `Integer` or `Boolean` type

vm/range_test.go
```diff
@@ -294,3 +294,171 @@ func TestInclude(t *testing.T) {
...
+	for i, tt := range testsFail {
+		evaluated := vm.testEval(t, tt.input)
+		err, ok := evaluated.(*Error)
+		if !ok {
+			t.Errorf("Expect error. got=%T (%+v)", err, err)
+		}
+		if err.Message != tt.expected.Message {
+			t.Errorf("Expect error message \"%s\". got=\"%s\"", tt.expected.Message, err.Message)
+		}
+		vm.checkCFP(t, i, 1)
```
Maxwell-Alexius: Error Testing Checked

st0012: Ok, let's merge this

PR #227, opened 2017-07-07
Implement #225 Part 2

### LINKED ISSUE #225: Make sure cfp (call frame pointer) to be 0 when program exit.
This is also a tricky issue like #221. We must make sure when code execution ends the call frame pointer would be 0, because all frames should be popped.

However, this rule only applies to program that is successfully executed. If we detected an error the program would be stopped so call frames stack will remain its state.

st0012:
This PR applies vm's CFP check to all test cases and closes #225 

When normal program's execution ends, the cfp should be 0.
But if the program has error the cfp should stop changing because the program is stopped.

### DIFF
--- a/vm/range.go
+++ b/vm/range.go
@@ -271,10 +271,19 @@ func builtInRangeInstanceMethods() []*BuiltInMethodObject {
 						return newError("Step can't be negative")
 					}
 
-					for i := ran.Start; i <= ran.End; i += stepValue {
-						obj := t.vm.initIntegerObject(i)
-						t.builtInMethodYield(blockFrame, obj)
+					// range end must greater or equal than range start to execute the block
+					if ran.End >= ran.Start {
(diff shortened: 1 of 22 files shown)


vm/range.go
```diff
@@ -271,10 +271,19 @@ func builtInRangeInstanceMethods() []*BuiltInMethodObject {
...
+						for i := ran.Start; i <= ran.End; i += stepValue {
+							obj := t.vm.initIntegerObject(i)
+							t.builtInMethodYield(blockFrame, obj)
+						}
+
+						return ran
 					}
+
+					// if block is not used, it should be popped
+					t.callFrameStack.pop()
```
st0012: @goby-lang/maintainers  This bug fix is the reason I add the CFP check, because some issues can't be found by looking at the end result only.

PR #228, opened 2017-07-07
Refactor Object and Class

st0012:
Simplify object and class' design and try to follow Golang's convention like using getter instead of expose type.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -43,11 +43,6 @@ func (a *ArrayObject) toJSON() string {
 	return out.String()
 }
 
-// Class returns current object's class, which is RArray
-func (a *ArrayObject) Class() *RClass {
-	return a.class
-}
-
 // length returns the length of array's elements
 func (a *ArrayObject) length() int {
 	return len(a.Elements)
--- a/vm/array.go
+++ b/vm/array.go
@@ -9,7 +9,7 @@ import (
 // An array is a collection of different objects that are ordered and indexed.
 // Elements in an array can belong to any class.
 type ArrayObject struct {
-	Class    *RClass
+	class    *RClass
 	Elements []Object
 }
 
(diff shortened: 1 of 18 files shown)


vm/array.go
```diff
@@ -43,9 +43,9 @@ func (a *ArrayObject) toJSON() string {
 	return out.String()
 }
 
-// returnClass returns current object's class, which is RArray
-func (a *ArrayObject) returnClass() Class {
-	return a.Class
+// Class returns current object's class, which is RArray
+func (a *ArrayObject) Class() *RClass {
```
st0012: This is the `getter` method I am talking about.

vm/object.go
```diff
@@ -22,26 +22,30 @@ func (p *Pointer) returnClass() *RClass {
 
 // RObject represents any non built-in class's instance.
 type RObject struct {
-	Class             *RClass
+	*baseObj
+	InitializeMethod *MethodObject
+}
+
+type baseObj struct {
```
st0012: By embedding `baseObj`, every object can have consist attributes like class and instance variables.

vm/instruction.go
```diff
@@ -256,14 +256,14 @@ var builtInActions = map[operationType]*action{
...
 
-			method := &MethodObject{Name: methodName, argc: argCount, instructionSet: is, class: t.vm.builtInClasses["Method"]}
+			method := &MethodObject{Name: methodName, argc: argCount, instructionSet: is, baseObj: &baseObj{class: t.vm.builtInClasses[methodClass]}}
 
 			v := t.stack.pop().Target
 			switch self := v.(type) {
 			case *RClass:
 				self.Methods.set(methodName, method)
 			default:
-				self.returnClass().(*RClass).Methods.set(methodName, method)
```
st0012: This can also reduce some type assertions.

hachi8833:
Good refactoring!
Actually, yesterday I suppose that `RClass` might be redundant, but I was not confident about this...

vm/object.go
st0012: And from now on,  if objects' function use same attribute we just need to implement it on `baseObj` once. Just like `instanceVariableGet` and `instanceVariableSet`

st0012:
@hachi8833 You can just point out what you think should be improved 😄

PR #235, opened 2017-07-10
Refactor: arrange the order of functions in built-in classes for consistency

hachi8833:
I hope this helps us.

1. type definitions
2. init*Object()
3. init*Class()
4. class methods
5. instance methods
6. polymorphic helper functions
7. other helper functions

### DIFF
--- a/vm/error.go
+++ b/vm/error.go
@@ -35,6 +35,19 @@ const (
 	UnsupportedMethodError = "UnsupportedMethodError"
 )
 
+// Error ...
+type Error struct {
+	*baseObj
+	Message string
+}
+
+func initErrorObject(errorType *RClass, format string, args ...interface{}) *Error {
+	return &Error{
+		baseObj: &baseObj{class: errorType},
+		Message: fmt.Sprintf(errorType.Name+": "+format, args...),
+	}
+}
+
 func (vm *VM) initErrorClasses() {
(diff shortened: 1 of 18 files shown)


vm/error.go
```diff
@@ -35,6 +35,19 @@ const (
 	UnsupportedMethodError = "UnsupportedMethodError"
 )
 
+// Error ...
```
st0012: @hachi8833 Can you also help document Error struct? thanks 😁

vm/vm.go
```diff
@@ -25,7 +25,7 @@ type filename string
 type errorMessage string
 
 var standardLibraries = map[string]func(*VM){
-	"file":              initializeFileClass,
+	"file":              initFileClass,
 	"net/http":          initializeHTTPClass,
```
st0012: Can you also change these function names?

vm/error.go
hachi8833: I'll look into it!

vm/vm.go
hachi8833: I got it!

vm/error.go
hachi8833: I'd do this on other PR

st0012: ok, then I'll merge this first.

st0012:
@hachi8833 Thanks a lot!

Issue #233, opened 2017-07-08
Going back REPL history with ↑ arrow key


hachi8833:
Extracted from #177.

Command history with up arrow key is not working for now:

>> ^[[A^[[A

st0012:
@hachi8833 I have no idea how to implement this, if you know you can give it a try 😅

hachi8833:
Memo: looks like the simple https://github.com/bobappleyard/readline would help us.
This depends on external readline lib, which is also used in Ruby.

Also found https://github.com/chzyer/readline (pure readline written in Go) but looks a little bit overkill.

hachi8833:
I found that https://github.com/bobappleyard/readline uses cgo and looks not good for Goby.
(I think one of the advantage is C-lang dependent)

st0012:
This has been implemented by @hachi8833

PR #238, opened 2017-07-10
WIP: Introducing readline in REPL

hachi8833:
For #233, I tried for my interests https://github.com/chzyer/readline a little and looks very fine for improving REPL.

Currently, just a single-line execution is possible, but all readline features such as command history or autocomplete (`help` and `exit`) are availables.
I'm wondering how to write tests for igb.go.

[readline-multiline.go](https://github.com/chzyer/readline/blob/master/example/readline-multiline/readline-multiline.go) is also an example of multiline execution. 

Could you take a look at this?

### DIFF
--- a/goby.go
+++ b/goby.go
@@ -25,7 +25,7 @@ func main() {
 	flag.Parse()
 
 	if *interactiveOptionPtr {
-		igb.Start(Version)
+		igb.StartIgb(Version)
 		os.Exit(0)
 	}
 
--- a/goby.go
+++ b/goby.go
@@ -1,17 +1,17 @@
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


st0012:
@hachi8833 Wow it seems that `readline` covers a lot implementation details for us. If this works well I think we don't need test for it or just write test script for CI to run.

st0012:
And can you open a branch on Goby for this (and the future works)? Because it would be more convenient if others needs to try/improve your code.

hachi8833:
Implemented multiple-line execution.
Though it is not complete yet, but works in many cases.
I'd open PR when completed.

PR #242, opened 2017-07-10
Fix while statement might cause stack overflow issue

### LINKED ISSUE #221: Instruction generation issue
Currently the code generator doesn't pop unused expression values. So program like this:

```ruby
100000.times do |i|
  1
end
```

will make value stack have 100000 integer 1.

But in test we need program to return the last expression to compare result, which means the code generator needs to be smart enough to use `pop` instruction to omit unused values and preserve the last pop before main program exit.

st0012:
I thought the issue described in #221 is small but it's not. This PR can only make compiler remove unused expressions. But there still are some instruction generation issue that will cause stack overflow condition.

This PR did following changes:

### Make postfix expressions return nothing

`a++` or `a--` should just change its value (BTW current implementation is not good) and shouldn't return anything. So I'll pop the value after updated variable. Actually I'm also thinking about removing these methods 🤔

### Add Stack Pointer checking function

Just like `checkCFP`

### Remove unused expressions declared in certain places

For example, these expression should be ignored when show up alone like:

```ruby
a
```

```ruby
1 + a
```

```ruby
Foo
```

Because they are useless and will keep stack growing unnecessarily. (They won't be popped to variable storage or popped by instruction so they'll just stay there.)

Following expressions should be removed when declared but not used

- Variable expressions like identifier, instance variable or constant
- Data type expressions like string, integer, array...etc.
- Self expression
- Prefix expression like !true or -5
- Not assignment infix expressions like: 1 + a * 5

But only when those they are inside following places:
- block argument
- method definition
- if expression's consequence or alternative block
- while statement

### If expression shouldn't returns nil if it's not assigned in some cases

In top level if expression returns nil if there's no alternative condition. This shouldn't cause much problem.
```ruby
if false
  x = 1
end # This pushes nil

x # This pushes nil too
```

But in some places like method definition, it can't returns `nil` or the stack will grow with `nil` if that method called multiple times.

```ruby
def foo
  if false
    x = 1
  end # This shouldn't push nil
end

foo # This should push nil
```


hachi8833:
Looks very hard... I'm unsure if I grasp the key of the issues.

At least, it looks like `++` and `--` operators does not fit pure OOPs language, so I suppose replacing to `+=` and `-=` is better.

FYI: this might help you https://github.com/pocke/rpr to get AST easily from current Ruby.

st0012:
@hachi8833 Let me explain a little more, for example:
Before this PR, the following code
```ruby
	i = 0

	while i < 100 do
	  10
	  i++
	end

	i
```

generates

```
0 putobject 0
1 setlocal 0 0
2 jump 9
3 putnil
4 pop
5 jump 9
6 putobject 10 <- look at here
7 getlocal 0 0
8 send ++ 0
9 getlocal 0 0
10 putobject 100
11 send < 1
12 branchif 6
13 putnil
14 pop
15 getlocal 0 0
16 leave
```

Which means every time we run the code block, it pushes a 10 on to the stack. But it's actually useless and will never be popped, which means stack never shrink.
So we need to ignore those expressions when generating instructions.

Issue #194, opened 2017-07-02
Higher unicode characters are garbled on `String#capitalize`


hachi8833:
```ruby
a = "漢"
a.capitalize # => "Æ��"  (Ruby returns "漢" as is)

b = "🍣"
b.capitalize  # => "Ð���" (Ruby returns "🍣" as is)
```

FYI: Goby's `upcase` and `downcase` works fine (= does not change such chars).

```ruby
a.upcase   #=> "漢"
a.downcase #=> "漢"
b.upcase   #=> "🍣"
b.downcase #=> "🍣"
```

hachi8833:
`reverse` causes the similar issue: looks like this works byte-based.

```ruby
a.reverse
#=> "¢¼æ"
```

hachi8833:
I suppose we should also change string type in Golang to rune type to handle multilingual characters.

st0012:
@Maxwell-Alexius @hachi8833 Has this been fixed?

Maxwell-Alexius:
Not yet, this needs another PR.

Maxwell-Alexius:
Close this issue with PR #254 merged

Issue #174, opened 2017-06-25
Goby's String class should be multilingual-conscious


hachi8833:
Current strings are counted as byte:

```ruby
puts("🍣".size)   # => 4
puts("我学".size) # => 6
puts("🍣".length)   # => 4
puts("我学".length) # => 6
```

This comes from the `string` type in Go.

```go
type StringObject struct {
	Class *RString
	Value string
}
```

I'd suggest that changing `string` type to `rune` type or like that as a default, to support `each` on multilingual strings. 

------------
**memorandum:** Looks like the minimum specification of Goby is needed soon: 

- Goby's encoding is implicitly fixed to UTF-8, while Ruby's String has `#encoding` to change with. 
- Perhaps character literal like `"\u{00A0}"` or numeric literal like `1_000_000` are going to be needed in the future, and such literals would affect lexer/parser implementations. 
- Implementing without specification would lead inconsistency among implementations.

FYI: Ruby had long been missing the official specification until ISO/IEC 30170:2012 was established https://www.iso.org/standard/59579.html.
The latest specification http://kikakurui.com/x3/X3017-2013-01.html is unfortunately written in Japanese.

Maxwell-Alexius:
Yeah, I do found out that `rune` type of string did help tackling some sort of character problems. When printing length of string, it counts the `bytes` of the string. However, `rune` type determines the character by its *codepoint*. See more on [this article](https://blog.golang.org/strings)

```
package main

import "fmt"

func main() {
    const sample = "🍣"
    fmt.Println(len(sample)) // => 4
    fmt.Println(len([]rune(sample)))  // => 1
}
```

I do think we can convert `string` to `[]rune` type but not for all of the string api methods. Maybe we can *correct the methods that needs to have precise conversion* which involves with **string indexing operation**.

For instance, `String#insert` need to specify the index of the string to insert an input string. However, `String#include` might not need to do the conversion because it just need to match the bytes of the string but not counting or modifying the index of the string.

Also, we can add multilingual or other bizarre unicode character or string to test the string methods. This deserves another PR.

Maxwell-Alexius:
Close this issue with PR #254 merged

PR #254, opened 2017-07-12
#174 & #194 Support UTF-8 Encoding in String Method API

Maxwell-Alexius:
This PR aims to support the details stated in #174 and #194.

For future string API optimizations, it deserves another PR. So it is very welcome to help enhance or refactor String API methods!

For now, Goby lang support multilingual by using plain `rune` and `string` conversion.

### DIFF
--- a/vm/string.go
+++ b/vm/string.go
@@ -288,6 +288,8 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 			// "Hello\nWorld"[5] # => "\n"
 			// "Hello"[-1]       # => "o"
 			// "Hello"[-6]       # => nil
+			// "Hello😊"[5]      # => "😊"
+			// "Hello😊"[-1]     # => "😊"
 			// ```
 			//
 			// @return [String]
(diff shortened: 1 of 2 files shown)


vm/string.go
```diff
@@ -288,6 +288,8 @@ func builtinStringInstanceMethods() []*BuiltInMethodObject {
 			// "Hello\nWorld"[5] # => "\n"
 			// "Hello"[-1]       # => "o"
 			// "Hello"[-6]       # => nil
+			// "Hello😊"[5]      # => "😊"
+			// "Hello😊"[-1]     # => "😊"
```
hachi8833: Suggestion: checking '4-byte UTF-8 codepoints' such as '🍣 ' or '🍺' as well is better.
Ref: http://mysqlserverteam.com/sushi-beer-an-introduction-of-utf8-support-in-mysql-8-0/

Maxwell-Alexius: Got it!

st0012:
I'm not very familiar with encoding, so I just took a look at test cases and this LGTM. @hachi8833 you can merge it when you think it's ready.

hachi8833:
As far as I see, `<=>` [line 225](https://github.com/Maxwell-Alexius/goby/blob/a9c82b93b8266c4c63d9fe3b1c262bb633012be9/vm/string.go#L225) should also be conscious of rune.

LGTM for the rest. 😄

hachi8833:
I noticed that I can merge and then Maxwell can fix `<=>` after that.
Merging...

Issue #247, opened 2017-07-11
Implement Hash API


HexterCH:
Import simple [Hash API](https://docs.ruby-lang.org/en/2.0.0/Hash.html)

- [x] each_key
- [x] each_value

st0012:
@HexterCH Implement, not import

Maxwell-Alexius:
More Hash API list from [Ruby Ver. 2.4.1](https://ruby-doc.org/core-2.4.1/Hash.html) :

- [ ] `Hash#delete` Should input `string` type or `symbol` type key
- [ ] `Hash#delete_if` or `Hash#reject` Pass in a proc which determines to delete the key or not
- [ ] `Hash#keep_if` or `Hash#select` Contrast to `delete_if` method
- [ ] `Hash#key?` or `Hash#has_key?` or `Hash#member?` ` @return [Boolean]`
- [ ] `Hash#value?` or `Hash#has_value?` `@return [Boolean]`
- [ ] `Hash#empty?` or `Hash#any?` `@return [Boolean]`
- [ ] `Hash#clear @return [Hash]`
- [ ] `Hash#compact @return [Hash]`, e.g. `{a: 1, b: "Hello", c: nil}.compact # => {a: 1, b: "Hello"}`
- [ ] `Hash#eql? @return [Hash]` (This might be a little big challenging to implement)
- [ ] `Hash#flatten @return [Array]`, e.g. `{a: 1, b: [2, "Hello"], c: nil}.flatten # => [:a, 1, :b, [2, "Hello"], :c, nil]`
- [ ] `Hash#to_s @return [String]`
- [ ] `Hash#to_a @return [Array]`
- [ ] `Hash#invert @return [Hash]`, e.g. `{a: 1, b: "Hello", c: nil}.invert # => {1 => :a, "Hello" => :b, nil => :c}`
- [ ] `Hash#keys @return[Array]` Return array of keys
- [ ] `Hash#values @return[Array]` Return array of values
- [ ] `Hash#length` or `Hash#size` `@return [Integer]`
- [ ] `Hash#merge @return [Hash]` Merging two hash, e.g. `{a: 1, b: "Hello"}.merge({c: "World"}) # => {a: 1, b: "Hello", c: "World"}`
- [ ] `Hash#replace @return [Hash]`, alike `String#replace` method
- [ ] `Hash#shift` Shift out a key-value pair, alike `Array#shift` method, this method may also be challenging to implement
[...]

PR #267, opened 2017-07-15
#247 Hash API Methods Implementation

Maxwell-Alexius:


vm/hash.go
```diff
@@ -231,6 +635,59 @@ func (h *HashObject) length() int {
...
+func (h *HashObject) sortedKeys() []string {
+	var arr []string
+	for k := range h.Pairs {
+		arr = append(arr, k)
+	}
+	sort.Strings(arr)
+	return arr
+}
+
+func (h *HashObject) isEqualTo(compare *HashObject) bool {
```
Maxwell-Alexius: Note: I think this has a better implementation

st0012: Can we use [reflect.DeepEqual](https://golang.org/pkg/reflect/#DeepEqual) in this case?

Maxwell-Alexius: Let me try it ~ I'll also look for the information about reflect in Go

vm/hash.go
```diff
@@ -133,24 +134,267 @@ func builtinHashInstanceMethods() []*BuiltInMethodObject {
...
+			// @return [Array]
+			Name: "each_key",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+					if len(args) != 0 {
+						return initErrorObject(ArgumentErrorClass, "Expect 0 argument. got=%d", len(args))
+					}
+
+					if blockFrame == nil {
+						t.returnError("Can't yield without a block")
```
Maxwell-Alexius: @st0012 Would you suggest me to change it to something like this?

```
return initErrorObject(InternalError, "Can't yield without a block")
```

st0012: https://github.com/goby-lang/goby/commit/9a428b14be809d5c7a4e850a4a1ce04b566996f0
We should improve this method to make it receive error type argument instead

hachi8833:
Checked out `feature/issue-247`. LGTM.

Maxwell-Alexius:
I'll rebase the branch first

vm/thread.go
```diff
@@ -174,7 +174,8 @@ func (t *thread) evalMethodObject(receiver Object, method *MethodObject, receive
 
 // TODO: Use this method to replace unnecessary panics
 func (t *thread) returnError(msg string) {
-	err := &Error{Message: msg}
+	err := initErrorObject(InternalErrorClass, msg)
+	//err := &Error{Message: msg}
```
Maxwell-Alexius: Changed the `err` to the `initErrorObject` format

vm/hash.go
```diff
@@ -200,8 +689,12 @@ func (h *HashObject) toString() string {
 	var out bytes.Buffer
 	var pairs []string
 
-	for key, value := range h.Pairs {
-		pairs = append(pairs, fmt.Sprintf("%s: %s", key, value.toString()))
+	for _, key := range h.sortedKeys() {
+		if _, isString := h.Pairs[key].(*StringObject); isString {
```
st0012: I don't like that we need to check this for every pair. But I don't have better idea to improve this yet.

vm/hash_test.go
```diff
@@ -3,42 +3,650 @@ package vm
...
+		evaluated := vm.testEval(t, tt.input)
+		testArrayObject(t, i, evaluated, tt.expected)
+		vm.checkCFP(t, i, 0)
+	}
+}
+
+func TestHashEachKeyMethodFail(t *testing.T) {
+	vm := initTestVM()
+
+	testArgumentError := `
```
st0012: I think we better not writing test in this way. We should still uses table structure for this like
```go
tests := []struct{
    input string
    expectedErr string
    expectedCfp int
} { ..... }
```

Maxwell-Alexius: Okay, I will change this

vm/hash.go
Maxwell-Alexius: I think I'll mark this as a TODO

PR #278, opened 2017-07-17
REPL: Change `restart` command to `reset` for simplification

hachi8833:
* `restart` command to `reset`
* Just added the following fortune when starting/resetting/exiting REPL. I just wonder it is OK for you.

image

### DIFF
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -25,7 +25,7 @@ const (
 	interrupt = "^C"
 	exit      = "exit"
 	help      = "help"
-	restart   = "restart"
+	reset     = "reset"
 
 	readyToExec = "readyToExec"
 	Waiting     = "waiting"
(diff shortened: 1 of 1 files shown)


st0012:
@hachi8833 Haha it's pretty funny, LGTM. But I think it's better to separate those emojis into categories, and display them according to the command. For example, only put positive emoji when REPL started.

hachi8833:
Thanks!
It'd be better you modify the behavior as you like.

st0012:
You should make the final decision on this, it's just a suggestion 😁

hachi8833:
Perhaps I'd arrange in a bit unobtrusive way: showing this only once would be nicer.

hachi8833:
Modified the fortune a bit.
In addition, I made the package variables to func variables in REPL.

Issue #237, opened 2017-07-10
Unify Error Style Refactoring


Maxwell-Alexius:
Another proposal. For example, inspecting the `vm/array.go` file, you can see that there are several types of error format:

`1st style: vm/array.go:58`
```
Name: "[]",
Fn: func(receiver Object) builtinMethodBody {
    return func(t *thread, args []Object, blockFrame *callFrame) Object {
        if len(args) != 1 {
        return &Error{Message: "Expect 1 arguments. got=%d" + string(len(args))}
    }
```

`2nd style: vm/array.go:101`
```
Name: "[]=",
Fn: func(receiver Object) builtinMethodBody {
    return func(t *thread, args []Object, blockFrame *callFrame) Object {
  
        // First arg is index
        // Second arg is assigned value
        if len(args) != 2 {
            return newError("Expect 2 arguments. got=%d", len(args))
        }
```

`3rd style: vm/array.go:243`
```
Name: "count",
Fn: func(receiver Object) builtinMethodBody {
    return func(t *thread, args []Object, blockFrame *callFrame) Object {
        arr := receiver.(*ArrayObject)
        var count int
  
        if len(args) > 1 {
            return initErrorObject(ArgumentErrorClass, "Expect 1 argument, got=%v", len(args))
        }
```

So, I think the error style format should all based on the third format with specific error message, such that:

Error Type|Description|Example Error Message Format
---|---|---
`ArgumentError`|Wrong input argument count|`Expect 2 arguments. got=%v` or `Expect 1 argument. got=%v`
`TypeError`|Wrong argument type|`Expect argument to be Range. got=%T` or `Expect index to be Integer. got=%T`
`ArgumentError`|Incorrect range or scope|`Second argument must be larger than or equal to 0. got=%v` or `Index must be smaller than 5. got=%v`

And every method *should* check the input argument count even if it doesn't need to have any arguments.

Maxwell-Alexius:
Closed with #284 Merged

PR #284, opened 2017-07-18
Refactor error system

st0012:
Refer to #237.

The error style code  is formally formatted in VM part. The rules are listed below.

- Part of the `vm/vm.go`, `thread.go` & `vm/instruction.go` uses the `t.returnError` as the error function

- VM part now unified `&Error`, `wrongTypeError`, `newError` ... different various style of errors into `t.vm.initErrorObject`.

- The first params of `initErrorObject` function now modified to pass in Go `string` type value instead of `*RClass` type value. For instance, previous format:

```go
return initErrorObject(ArgumentErrorClass, "Expect 1 argument. got: %d", len(args))
```

should now become:

```go
return t.vm.initErrorObject(ArgumentError, "Expect 1 argument. got: %d", len(args))
```

- The error message will also abandon the pure code-in string format, instead using constants listed in `vm/error.go` to represent the error message. For instance, instead of code-in string format:

```go
return t.vm.initErrorObject(InternalError, "Can't yield without a block")
```

should now use:

```go
return t.vm.initErrorObject(InternalError, CantYieldWithoutBlockFormat)
```

- Error (or API failure) testing should not use the `*Error` format. To implement failure testing of error message, use pure Go `string` format, for instance:

```go

func TestArrayShiftMethodFail(t *testing.T) {
	testsFail := []struct {
		input    string
		expected string // Do not use *Error
	}{
		{`
		a = [1, 2]
		a.shift(3, 3, 4, 5)
		`, "ArgumentError: Expect 0 argument. got=4"}, // Error message using string format
	}

	for i, tt := range testsFail {
		v := initTestVM()
		evaluated := v.testEval(t, tt.input)
		// Using checkError function to check error message
		checkError(t, i, evaluated, ArgumentError, tt.expected)
		v.checkCFP(t, i, 1)
	}
}
```

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -59,14 +59,14 @@ func builtinArrayInstanceMethods() []*BuiltInMethodObject {
 			Fn: func(receiver Object) builtinMethodBody {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
 					if len(args) != 1 {
-						return &Error{Message: "Expect 1 arguments. got=%d" + string(len(args))}
+						return t.vm.initErrorObject(ArgumentError, "Expect 1 arguments. got=%d", len(args))
 					}
 
 					i := args[0]
 					index, ok := i.(*IntegerObject)
 
 					if !ok {
-						return newError("Expect index argument to be Integer. got=%T", i)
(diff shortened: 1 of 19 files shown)


Maxwell-Alexius:
LGTM

PR #291, opened 2017-07-20
Implement importing go package's mechanism

st0012:
# Purpose
This PR aims at making importing and calling function's in other Go package possible.

### Note
- `plugin` build mode is only supported on `linux` platform for now.
- `plugin` build mode can only compiles main package for now.

## Sample

Assume we have a third-party package like this in `github.com/goby-lang/plugin/plugin.go`:
```go
package main

type Bar struct {
	name string
}

func (b *Bar) Name() string {
	return b.name
}

func NewBar(name string) (*Bar, error) {
	return &Bar{name: name}, nil
}

func main() {
	fmt.Println("Main")
}
```

And we can import it and use it's exported functions and types like:

```ruby
p = import "github.com/goby-lang/plugin/plugin.go"
result = p.send("NewBar", "xyz") # multiple result, so result is an array
bar = result[0] # The second element is error
bar.send("Name") #=> "xyz"
```

## Explanation

I added two built in classes for this feature 1) `Plugin` 2) `Struct`

- `Plugin` carries a pointer of [`plugin.Plugin`](https://golang.org/pkg/plugin/#Plugin)
- `Struct` stores go's `interface{}`, it's a Goby representation of any Go's object

And also added `Object#import`, which compiles the target file with `plugin` build mode and use `plugin.Open` to retrieve the plugin

So the flow of sample is

1. Use import to compile and get plugin object
2. Store retrieved `*plugin.Plugin` in a Goby's `Plugin` object
3. Call package's `NewBar` function with an argument `"xyz"`, which returns a pointer of `Bar` and error
4. Store `*Bar` and `error` in two `Struct` object
5. Call `*Bar`'s `Name` method and get its name, which is `"xyz"`.

### If the function have different type

For example: 

```go
func (b *Bar) Add(x int, y int64) int64 {
	return int64(x) + y
}
```

We should provide user type conversion methods on built in types (in this case is `Integer`)

So the Goby code would look like:

```ruby
bar.send("Add", 10, 100.to_int64)
```

## Other todos:

- Implement all type conversion methods
- Make sure we can also convert all Go types into corresponding Goby object
- Make `import` search share object file before building
- Refactor temp implementations in `test.sh` and make plugin and struct's test coverage can be counting too.

(Dockerfile is just for testing the program because the `plugin` build mode is only available on linux.)


hachi8833:
Well, how about a mechanism for locking and updating packages?

st0012:
@hachi8833 That's package management, which is a totally different issue. I think we can't finish that before RubyKaigi

PR #300, opened 2017-07-24
Fix unindentation behavior in REPL

### LINKED ISSUE #299: Unindentation might drop statements in REPL
The following comes from [statement_test](https://github.com/goby-lang/goby/blob/master/vm/statement_test.go#L273) and the result in REPL is different from normal execution:

```ruby
x = 0
y = 0

while x < 10 do
  x = x + 1
  if x == 5
    next
  end
  y = y + 1      # looks like the statement is missing.
end

x + y              #=> should be 19, actually 10
```

Changing `x` to `y` in `while` condition, the following goes infinite loop. 

```ruby
x = 0
y = 0

while y < 10 do
  x = x + 1
  if x == 5
    next
  end
  y = y + 1  
end          # infinite loop
```

Looks like the line `y = y + 1` is missing when unindenting.

hachi8833:
Critical. This should fix #299.

### DIFF
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -112,7 +112,6 @@ reset:
 				if len(line) == 0 && cmds == nil {
 					println("")
 					println("Bye!")
-					return
 				} else {
 					// Erasing command buffer
 					println("")
@@ -122,6 +121,7 @@ reset:
 					cmds = nil
 					continue
 				}
+				return
 			}
 		}
 
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -155,7 +155,6 @@ reset:
 				if !sm.Is(Waiting) {
 					sm.Event(Waiting)
 				}
-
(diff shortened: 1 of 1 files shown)


igb/repl.go
```diff
@@ -178,10 +172,25 @@ reset:
...
+				sm.Event(waitEnded)
+				rl.SetPrompt(prompt(stack) + indent(stack))
+				cmds = append(cmds, line)
+			}
 		}
 
-		if sm.Is(Waiting) {
+		if sm.Is(Waiting) && stack > 0 {
 			println(prompt(stack) + indent(stack) + line)
+			sm.Event(Waiting)
```
st0012: If it's already waiting, why make it waiting again?

hachi8833: Thank you for pointing this out.
I'd remove it.

Issue #241, opened 2017-07-10
Unsupported syntactic sugar


st0012:
- [x] `+=`
- [x] `-=`
- [x] `||=`

PR #305, opened 2017-07-25
#241 Basic Implementation for Assignment by Operator Syntax Suger

Maxwell-Alexius:


vm/class_test.go
```diff
@@ -578,6 +578,56 @@ func TestClassGeneralComparisonOperation(t *testing.T) {
...
+	}
+}
+
+func TestGeneralAssignmentByOperationFail(t *testing.T) {
+	testsFail := []struct {
+		input   string
+		errType string
+		errMsg  string
+	}{
+		{`a = 1; a ||=`, ArgumentError, "ArgumentError: Expect 1 argument. got: 0"},
```
st0012: This shouldn't be test here because it'll fail during parsing.

st0012: And you can implement a pattern for testing parsing error if you want to.

Maxwell-Alexius: Okay, I'll try it~

PR #308, opened 2017-07-26
REPL: Refactoring and small fix Igb

hachi8833:
The refactoring is for writing test, and under the way.
d258be7 is a small fix for indenting `#`.

### DIFF
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -38,6 +38,7 @@ const (
 	emojis = "😀😁😂🤣😃😄😅😆😉😊😋😎😍😘😗😙😚🙂🤗🤔😐😑😶🙄😏😮😪😴😌😛😜😝🤤🙃🤑😲😭😳🤧😇🤠🤡🤥🤓😈👿👹👺💀👻👽🤖💩😺😸😹😻😼😽"
 )
 
+// Igb holds internal states of Igb.
 type Igb struct {
 	sm        *fsm.FSM
 	rl        *readline.Instance
@@ -47,6 +48,7 @@ type Igb struct {
 	stack     int
 }
 
+// Ivm holds VM only for Igb.
 type Ivm struct {
 	v *vm.VM
 	p *parser.Parser
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -134,7 +134,8 @@ reset:
 		program, perr := ivm.p.ParseProgram()
 
 		if perr != nil {
-			if perr.IsEOF() {
+			switch {
+			case perr.IsEOF():
 				if !igb.sm.Is(Waiting) {
 					igb.sm.Event(Waiting)
 				}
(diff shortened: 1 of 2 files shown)


hachi8833:
Needs `git rebase master` first.

hachi8833:
Rebase is unnecessary. Reopened.

st0012:
@hachi8833 I think even if you need to rebase it doesn't mean you have to re-create the PR

igb/repl.go
```diff
@@ -38,172 +38,150 @@ const (
...
 )
 
+// Igb holds internal states of Igb.
+type Igb struct {
+	sm        *fsm.FSM
+	rl        *readline.Instance
+	completer *readline.PrefixCompleter
+	line      string
+	cmds      []string
+	stack     int
```
st0012: Actually I still don't understand what `stack` in igb means. I may confuse people since normally we think `stack` as a kind of data structure.

igb/repl.go
```diff
@@ -220,39 +198,90 @@ reset:
...
+			igb.cmds = nil
 		}
 	}
 }
 
 // Polymorphic helper functions --------------------------------------------
 
+// Other helper functions --------------------------------------------------
+
+func initIgb() *Igb {
```
st0012: usually it will be named as `NewIgb`

igb/repl.go
```diff
@@ -220,39 +198,90 @@ reset:
...
+		),
+		completer: readline.NewPrefixCompleter(
+			readline.PcItem(help),
+			readline.PcItem(reset),
+			readline.PcItem(exit),
+		),
+	}
+}
+
+func createVM() Ivm {
```
st0012: `NewVM`

igb/repl.go
```diff
@@ -38,172 +38,150 @@ const (
...
+	sm        *fsm.FSM
+	rl        *readline.Instance
+	completer *readline.PrefixCompleter
+	line      string
+	cmds      []string
+	stack     int
+}
+
+// Ivm holds VM only for Igb.
+type Ivm struct {
```
st0012: In the function (`createVM`) below you uppercased the `VM` word, why don't we also uppercase vm here? Normally in Golang it should be uppercased since it's an abbreviation, like JSON or HTML.

igb/repl.go
```diff
@@ -38,172 +38,150 @@ const (
 	emojis = "😀😁😂🤣😃😄😅😆😉😊😋😎😍😘😗😙😚🙂🤗🤔😐😑😶🙄😏😮😪😴😌😛😜😝🤤🙃🤑😲😭😳🤧😇🤠🤡🤥🤓😈👿👹👺💀👻👽🤖💩😺😸😹😻😼😽"
 )
 
+// Igb holds internal states of Igb.
+type Igb struct {
```
st0012: Are you planning use this struct outside this Package? If not you shouldn't export it (and `Ivm`).

igb/repl.go
st0012: I think this `stack` is more like a pointer?

igb/repl.go
hachi8833: I should've noticed that. Will make it private.

igb/repl.go
hachi8833: Yes. `indent` should be better in the case.

igb/repl.go
hachi8833: I will change this as well.

igb/repl.go
hachi8833: Got it!

igb/repl.go
hachi8833: I'll change this as well.

st0012:
@hachi8833 Can you rebase and merge it?

PR #315, opened 2017-08-02
[WIP] Regexp class is coming

hachi8833:
st0012 suggested me to implement Regexp with string literal first.
For now, only one class method and one instance method has been implemented. But it works!

The package github.com/dlclark/regexp has not been so visible, but is based on .NET Framework's powerful regexp library. Almost all equivalent to Onigmo, full-multilingual-conscious.

Thank you very much, st0012 and dlclark!


hachi8833:
Just to close to recreate WIP PR.

PR #326, opened 2017-08-11
Add flatten and join methods to Array class

shuheiktgw:
Hi, I just added flatten and join methods to Array class. I'm not that fluent with golang yet, so I would really appreciate it if you would comment whatever crosses your mind.

Thanks! 😄

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -66,6 +66,22 @@ func (a *ArrayObject) toJSON() string {
 	return out.String()
 }
 
+// flatten returns a array of Objects that is one-dimensional flattening of Elements
+func (a *ArrayObject) flatten() []Object {
+	var result []Object
+
+	for _, e := range a.Elements {
+		arr, isArray := e.(*ArrayObject)
+		if isArray {
+			result = append(result, arr.flatten()...)
+		} else {
+			result = append(result, e)
+		}
+	}
+
+	return result
+}
+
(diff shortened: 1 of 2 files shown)


vm/array.go
```diff
@@ -473,6 +489,70 @@ func builtinArrayInstanceMethods() []*BuiltInMethodObject {
...
+			// c = [ a, b, 9, 10 ] # => [[1, 2, 3], [4, 5, 6, [7, 8]], 9, 10]
+			// c.flatten # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+			// ```
+			// @return [Array]
+			Name: "flatten",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+					arr := receiver.(*ArrayObject)
+
+					if len(args) != 0 {
```
shuheiktgw: I know ruby's `flatten` method can take an argument: level. But I thought I do not have to implement that feature right now (since I myself never ran across the situation I have to give level to flatten.).

If you think that is necessary, I'm happy to work on it! 👍

st0012: I don't think it' necessary either

st0012:
@shuheiktgw Thank you! It LGTM, I'm going to merge it.

Issue #328, opened 2017-08-11
Support elsif keyword


st0012:

st0012:
@shuheiktgw This task only touch the compiler part for generating right instructions. You should be familiar with this since you also took nand2tetris 😆

st0012:
This has already been implemented by @shuheiktgw

PR #332, opened 2017-08-15
Add elsif keyword support

shuheiktgw:
This relates to https://github.com/goby-lang/goby/issues/328.
I added `elsif` support in for goby!

Thanks! 😄


compiler/ast/expressions.go
```diff
@@ -230,6 +235,28 @@ func (ie *IfExpression) String() string {
 	return out.String()
 }
 
+type ConditionalExpression struct {
```
st0012: @shuheiktgw Do you think we can use same struct to implement `case` statement?

shuheiktgw: Oh, yes! I've never thought about it but now that you mention it, I think we can use this for case statement since ultimately ruby"s case statement is just a series of if, elsif and else statements! 😄 

Did you already support case statement? If not, I'm glad to work on it after finishing elsif 👍

st0012: No I haven't support that yet

shuheiktgw: Ok, then let me work on it after finishing this! 😄

compiler/bytecode/expression_generation_test.go
```diff
@@ -71,6 +71,62 @@ func TestIfExpressionWithoutAlternativeCompilation(t *testing.T) {
 	compareBytecode(t, bytecode, expected)
 }
 
+func TestIfExpressionWithElsifCompilation(t *testing.T) {
```
shuheiktgw: I can vaguely understand what the  vm codes are trying to do, still I"m not that confident with it. So, I would really appreciate it if you review here a bit more carefully than the other parts I wrote...

shuheiktgw:
I finished working on elsif support! So could you please review this PR?
Thanks! 👍

st0012:
@shuheiktgw LGTM, nice job!

shuheiktgw:
Thanks!

PR #340, opened 2017-08-17
Add Doc to String#fmt and DB standard library

Maxwell-Alexius:


st0012:
@Maxwell-Alexius LGTM. Can you also add these document into gitbook's DB part?

Maxwell-Alexius:
I'll add it later on~

PR #341, opened 2017-08-18
Update README

hachi8833:
Restructured. Still updating a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,23 +7,218 @@
 [![codecov](https://codecov.io/gh/goby-lang/goby/branch/master/graph/badge.svg)](https://codecov.io/gh/goby-lang/goby)
 [![Readme Score](http://readme-score-api.herokuapp.com/score.svg?url=goby-lang/goby)](http://clayallsopp.github.io/readme-score?url=goby-lang/goby)
 
-**Join us on Slack!** [![](https://goby-lang-slackin.herokuapp.com/badge.svg)](https://goby-lang-slackin.herokuapp.com)
+**Goby** is an object-oriented interpreter language deeply inspired by **Ruby** as well as its core implementation by 100% pure **Go**, plus standard libraries to several features such as Plugin system. Note that we do not intend to reproduce whole of the honorable works of Ruby syntax/implementation/libraries. 
 
-Goby is an object-oriented interpreter language deeply inspired by Ruby and written in 100% pure Go. The goal of Goby is to help web developers create api servers or microservices simply and efficiently, with a help of tough thread-mechanism from Go's goroutine, see the [thread's example](https://github.com/goby-lang/goby/blob/master/samples/one_thousand_threads.gb). Howerver, We do not intend to reproduce all of works in Ruby implementation.
+One of our goal is to provide web developers a sort of small and handy environment that mainly focus on creating **API servers or microservices**. For this, Goby includes the following native features:
 
-Goby will finally equip a reduced set of Ruby's fundamental syntax, including Ruby's common methods and libraries, but will not equip most of Ruby's meta-programming magic to make Goby VM simpler. Goby will also finally equip a built-in HTTP library and multi-threaded server that comes from Go's HTTP packages.
+- Tough thread/channel mechanism powered by Go's goroutine
+- Builtin high-performance HTTP server
+- Database adapter (currently for PostgreSQL)
+- JSON support
+- Plugin system that can load existing Go packages dynamically and access from Goby
+    - Only for Linux by now.
 
-**Checkout our [sample app](http://sample.goby-lang.org) built with Goby! And [here](https://github.com/goby-lang/sample-web-app) is the source code**
+## Table of contents
+
+- [Demo and sample Goby app](#demo_and_sample_app)
+- [Aspects](#aspects)
+    - [Features](#features)
+    - [Language](#language)
+    - [Native class](#native_class)
+    - [Standard class](#standard_class)
+- [Installation](#installation)
+- [Usage](#usage)
+- [Sample codes](#sample_codes)
+- [Documentation](#documentation)
+- [Joining to Goby](#joining-to-goby)
+- [Maintainers](#maintainers)
+- [Support us](#support-us)
+- [References](#references)
 
 Goby interpreter is a monolithic binary executable, which consists of a YARV-like VM and a compiler, and a REPL (which is better than `irb`!). All components of Goby compiler, such as AST, lexer, parser, token, are written in 100% pure Go, instead of using conventional static yacc/lex/bison conversion. Goby maintainers don't need to care about C language anymore!
 
-We are optimizing and expanding Goby all the time and need your help. One of our vision is to utilize and manages tons of Go's packages easily from Goby scripts. Currently we already have some progress, see the introduction of Goby's [Plugin System](https://goby-lang.gitbooks.io/goby/content/plugin-system.html)
-
-[**User Manual (WIP)**](https://goby-lang.gitbooks.io/goby/content/)
-
-**Demo:**
+## Demo screen and sample Goby app
 
 <img src="http://i.imgur.com/5RxFgIW.gif?1" width="60%">
+
+**New!** Check-out our [sample app](http://sample.goby-lang.org) built with Goby. Source code is also available [here](https://github.com/goby-lang/sample-web-app).
+
+## Aspects
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,208 +7,194 @@
...
 
-Become a sponsor and get your logo on our README on Github with a link to your site. [[Become a sponsor](https://opencollective.com/goby#sponsor)]
+**Class**: Single inheritance, and 'module' is supported for mixin. Goby has internally two kinds of class: native class and standard class. **Native class** (or builtin class) provides fundamental methods. Most common methods such as `#puts` are in `Class` class. `Object` class is just a root class for consistency and holds no methods. **Standard class** (or standard library) can be loaded via `require` and provides additional methods. Standard classes are often split internal Go code and external Goby code in order to make implementation easier. Both kind of classes are transparent to Goby developers and can be overridden by child classes. Goby's Class hierarchy is very simple: `Object::Class::(anything like that)`.
```
st0012: I'm not sure what do you mean about the Class hierarchy. All constants is under `Object` namespace, but not necessarily under `Class` namespace.

st0012: I think you can mention we also implemented singleton classes in this section.

README.md
```diff
@@ -7,208 +7,194 @@
...
+ 
+Written in Go.
+
+- Class
+- Integer
+- String
+- Boolean
+- Null (nil)
+- Hash (with `#to_json` method)
+- Array (with `#each`)
```
st0012: I think we don't have to specify `Array` and `Range` have `#each` method. It's quite basic.
But if we implemented `enumerable` that'd be something we can talk about.

README.md
```diff
@@ -7,208 +7,194 @@
...
+- Range (with `#each`)
+- URI
+- Channel
```
st0012: There's a missing class: `GoObject`. Which stores any type of data that can't be transform to Goby's type directly. And we can use `GoObject#go_func` method to call the Go object's function.

README.md
```diff
@@ -7,208 +7,194 @@
...
+- [Joining to Goby](#joining-to-goby)
+- [Maintainers](#maintainers)
+- [Support us](#support-us)
+- [References](#references)
 
-Goby interpreter is a monolithic binary executable, which consists of a YARV-like VM and a compiler, and a REPL (which is better than `irb`!). All components of Goby compiler, such as AST, lexer, parser, token, are written in 100% pure Go, instead of using conventional static yacc/lex/bison conversion. Goby maintainers don't need to care about C language anymore!
+## Demo (REPL: igb)
 
-We are optimizing and expanding Goby all the time and need your help. One of our vision is to utilize and manages tons of Go's packages easily from Goby scripts. Currently we already have some progress, see the introduction of Goby's [Plugin System](https://goby-lang.gitbooks.io/goby/content/plugin-system.html)
+<img src="http://i.imgur.com/5RxFgIW.gif?1" width="60%">
```
st0012: BTW I think we should record a new demo. Maybe something like open REPL and insert/query database?

README.md
```diff
@@ -7,208 +7,194 @@
...
-<a href="https://opencollective.com/goby/backer/26/website" target="_blank"><img src="https://opencollective.com/goby/backer/26/avatar.svg"></a>
-<a href="https://opencollective.com/goby/backer/27/website" target="_blank"><img src="https://opencollective.com/goby/backer/27/avatar.svg"></a>
-<a href="https://opencollective.com/goby/backer/28/website" target="_blank"><img src="https://opencollective.com/goby/backer/28/avatar.svg"></a>
-<a href="https://opencollective.com/goby/backer/29/website" target="_blank"><img src="https://opencollective.com/goby/backer/29/avatar.svg"></a>
+Goby has several aspects: language specification, design of compiler and vm, implementation (just one for now), library, and the whole of them. 
 
+----------
 
-### Sponsors
+**Language**: Class-based, straignt-ahead object-oriented script language. Syntax is influenced by Ruby language (and by Go a bit), but has been **reduced and simplified** (and slightly modified) to keep Goby VM simple and concise. Some aspects of Ruby such as meta-programming (known as 'magic'), special variables with `$`, or like that, have been dropped.
```
st0012: I'd like to add a note here indicate that although these features are dropped for now but might come back later with different form or implementation.

README.md
```diff
@@ -7,208 +7,194 @@
...
 
-<img src="http://i.imgur.com/5RxFgIW.gif?1" width="60%">
+One of our goal is to provide web developers a sort of small and handy environment that mainly focus on creating **API servers or microservices**. For this, Goby includes the following native features:
 
-Support us with a monthly donation and help us continue our activities. [[Become a backer](https://opencollective.com/goby#backer)]
+- Tough thread/channel mechanism powered by Go's goroutine
+- Builtin high-performance HTTP server
+- Database adapter (currently for PostgreSQL)
+- JSON support
+- Plugin system that can load existing Go packages dynamically and access from Goby
```
st0012: We need to notify users this is only supported on Linux for now.

README.md
hachi8833: Sure, I'll do later!

README.md
hachi8833: Got it

README.md
hachi8833: Got that as well

README.md
hachi8833: Class hierarchy has just come from my question and noted temporary for clarification. 
Perhaps we should move some to around Gitbook.
I will revise the paragraph, adding singleton class.

README.md
hachi8833: Got it. 
(Looks I was thrown off from the high-speed development)

README.md
```diff
@@ -7,210 +7,206 @@
...
+### B. From Source
 
-**If you have any issue installing Goby, please open an issue for it**
+Try this if you'd like to contribute Goby! Skip 1 if you already have Golang in your environment.
 
-## Usage
+1. Prepare Golang environment
+    a. Install Golang
+    b. Make sure `$GOPATH` in your shell's config file( like .bashrc) is correct
+    2. Add you add your `$GOPATH/bin` to `$PATH`in your shell's env var
```
st0012: Here seems to have abnormal indentation. And the third point is missing

README.md
```diff
@@ -7,210 +7,206 @@
...
+- Integer
+- String
+- Boolean
+- Null (nil)
+- Hash
+- Array
+- Range
+- URI
+- Channel
+- GoObject (provides `#go_func`)
```
st0012: GoObject holds pure Go objects(or pointers) and  provides `#go_func` to interact with them

README.md
```diff
@@ -7,210 +7,206 @@
...
+
+- Loadable class
+    - `File`
+    - `DB` (only for PostgreSQL by now)
+    - `Plugin`
+- Loadable module
+    - NET
+        - `Net::HTTP:Request`
+        - `Net::HTTP:Response`
+        - `Net::SimpleServer` (See the performance by runniing [sample](https://github.com/goby-lang/goby/blob/master/samples/server.gb)!)
```
st0012: We should also put the sample site's link here

README.md
```diff
@@ -7,210 +7,206 @@
...
 
-    ```
-    $ go get github.com/goby-lang/goby
-    ```
-5. Set `GOBY_ROOT` to the project's root path, which should be:
+## Installation
```
st0012: I want to add a `using docker` section in here. I'll provide link to the docker repository, basic usage and extended usages.

README.md
hachi8833: Added

README.md
hachi8833: Thanks! I'm unable to use Docker so I'd be glad if you add them here after that.

README.md
hachi8833: Fixed

README.md
```diff
@@ -7,210 +7,206 @@
...
 
-**If you have any issue installing Goby, please open an issue for it**
+Try this if you'd like to contribute Goby! Skip 1 if you already have Golang in your environment.
 
-## Usage
+1. Prepare Golang environment
+    1-1. Install Golang
+    1-2. Make sure `$GOPATH` in your shell's config file( like .bashrc) is correct
+    1-3. Add you add your `$GOPATH/bin` to `$PATH`in your shell's env var
+4. Run `go get github.com/goby-lang/goby`
```
st0012: This should be 2?

hachi8833: I'll fix this

PR #361, opened 2017-08-25
Upgrade go to 1.9

st0012:
First thing to do is to use concurrent map.


st0012:
It turns out that TravisCI hasn't support `1.9` yet

st0012:
Skip `codeclimate` check because it hasn't support Go 1.9

Issue #355, opened 2017-08-24
Need methods to escape characters 


st0012:
like `"foo <span style="color:red">bar</span>"` should be `"foo <span style=\"color:red\">bar</span>"`

Maxwell-Alexius:
This is the log in interactive Goby, seems the `\"` has some problem which I never expected :

```
Goby 0.1.0 🤡 😮 🤡
» "a"
#» a
» "\"a\""
unexpected \ Line: 0
» "\"a\""
»  
```

```
Goby 0.1.0 😗 😀 😸
» { hello: "world" }
#» { hello: "world" }
» { hello: "world" }.to_json
#» {"hello":"world"}
» { hello: "\"world\"" }
expected next token to be }, got IDENT instead. Line: 0
» { hello: "\"world\"" }
» { hello: "\"world\"" }.to_json
expected next token to be }, got IDENT instead. Line: 0
» { hello: "\"world\"" }.to_json
»  
```

st0012:
@Maxwell-Alexius Can this be closed?

Maxwell-Alexius:
Yep! Thanks Stan!

PR #366, opened 2017-08-26
#355 Fix String Excaped Character

Maxwell-Alexius:


Maxwell-Alexius:
This PR aims to fix #355

st0012:
@Maxwell-Alexius Can you also add vm tests (just like lexer's)? Because this string might be changed during this process `AST->bytecode->VM's value`

Maxwell-Alexius:
Is it sth like string test which have escaped characters?

_Sent from my Htc HTC_D728x using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

st0012:
@Maxwell-Alexius Yes, it's better to add new test cases to every string function tests. And you don't have to add every escape cases I think, just focus on `\"` and `\'`

Maxwell-Alexius:
Updated!

PR #368, opened 2017-08-27
Forbidden duplicated method parameters

shuheiktgw:
I just noticed goby allows to define methods with the same argument name, which is prohibited in ruby. An example below

```
def someMethod(a, a)
  return a
end
```

So I just disabled it! Thanks! 😄

### DIFF
--- a/compiler/parser/statement_parsing.go
+++ b/compiler/parser/statement_parsing.go
@@ -114,6 +114,10 @@ func (p *Parser) parseParameters() []ast.Expression {
 		p.nextToken()
 		p.nextToken()
 		param := p.parseExpression(NORMAL)
+
+		if contains(params, param) {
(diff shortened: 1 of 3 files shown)


shuheiktgw:
I just finished implementing it, so could anyone review this PR?

compiler/parser/statement_parsing.go
```diff
@@ -117,6 +117,10 @@ func (p *Parser) parseParameters() []ast.Expression {
 		p.nextToken()
 		p.nextToken()
 		param := p.parseExpression(NORMAL)
+
+		if containsArg(params, param) {
```
st0012: I recommend renaming it to `paramDuplicated` which will be more clear as a condition.

st0012:
I didn't notice this, thanks! And remember to fix codeclimate's warning.

shuheiktgw:
Oops, sorry I just fixed the problems mentioned above!
Thanks for pointing out! 👍

PR #371, opened 2017-08-28
POST request functionality to Net::HTTP class and tests for .get() and .post() methods

ear7h:
http.go
	HTTP.post() - uses native go http.Post() method to post data via http
http_test.go tests for http class
	HTTP.get()
	HTTP.post()

### DIFF
--- a/vm/http.go
+++ b/vm/http.go
@@ -3,6 +3,8 @@ package vm
 import (
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"path"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


vm/http_test.go
```diff
@@ -0,0 +1,66 @@
...
+		Net::HTTP.get("http://127.0.0.1:3000")
+		`, "GET Hello World"},
+		{`
+		require "net/http"
+
+		Net::HTTP.post("http://127.0.0.1:3000", "text/plain", "Hi Again")
+		`, "POST Hi Again"},
+	}
+
+	c := make(chan bool, 1)
```
st0012: I want to move the whole server setup part above the test cases.

vm/http_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+		m.HandleFunc("/", func (w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+
+			if r.Method == http.MethodPost {
+				b, err := ioutil.ReadAll(r.Body)
+				if err != nil {
+					panic(err)
+				}
+				w.Write([]byte(r.Method + " " + string(b)))
```
st0012: I think just write `fmt.Sprintf("POST %s", string(b))` is better. Because we already know this method is surely `POST`

vm/http_test.go
```diff
@@ -0,0 +1,66 @@
...
+			} else {
+				w.Write([]byte(r.Method + " Hello World"))
```
st0012: Same as above

ear7h: I think using r.Method on this line is preferable as it can be used to test DELETE, HEAD, etc.

vm/http_test.go
ear7h: That's true I'll change that

vm/http_test.go
st0012: I see, I didn't know you will also add those methods.

ear7h:
Committed changes from your feedback

st0012:
One more thing, can you also add tests for false usage and none successful responses?
Those should be placed at new test functions, you can see error tests at array.go. Find those function which's name contains "Fail".

ear7h:
also changed the way url's are connected for get requests

vm/http.go
```diff
@@ -54,23 +57,64 @@ func builtinHTTPClassMethods() []*BuiltInMethodObject {
...
+
+					body := args[2].(*StringObject).value
+
+					resp, err := http.Post(uri.String(), contentType, strings.NewReader(body))
 
 					if err != nil {
 						return t.vm.initErrorObject(InternalError, err.Error())
 					}
+					if resp.StatusCode != http.StatusOK {
+						return t.vm.initErrorObject(InternalError, resp.Status)
```
st0012: BTW This error can be more specific, like `HTTPError`. You can create new error type in `error.go`. But let's merge this first.

st0012:
Because codeclimate still haven't upgrade its `golint`, so I still need to skip the codeclimate check.

PR #369, opened 2017-08-28
colors on Windows

mattn:
* colors on Windows
* strict path for temporary directory
* disable emoji(s)

windows

PR #381, opened 2017-08-31
colors on Windows by @mattn

st0012:
This PR is related to #369 and for helping mattn's branch install dependency.


st0012:
@mattn Looks like it works now, I'll merge this and closes #369. Thanks for your help!

PR #384, opened 2017-09-01
Refactor method parameter parsing

st0012:
- Consider splat argument as a type of prefix expression
- Restrict parameters define order: normal arg > optioned arg > splat arg
- Forbidden splat arg be defined twice

### DIFF
--- a/compiler/parser/statement_parsing.go
+++ b/compiler/parser/statement_parsing.go
@@ -84,33 +84,10 @@ func (p *Parser) parseDefMethodStatement() *ast.DefStatement {
 		p.nextToken()
 
 		switch p.peekToken.Type {
-		// def foo()
 		case token.RParen:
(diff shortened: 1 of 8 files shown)


compiler/parser/statement_parsing.go
```diff
@@ -137,25 +114,70 @@ func (p *Parser) parseParameters() []ast.Expression {
...
 		params = append(params, param)
 	}
 
+	p.checkMethodParameters(params)
+
 	p.fsm.Event(backToNormal)
 	return params
 }
 
+func (p *Parser) checkMethodParameters(params []ast.Expression) {
```
st0012: @shuheiktgw I added a function for checking parameters, currently only checking their order and duplication.

ear7h:
I agree with the design choices, I think it'll lead to faster execution and more organized code.

PR #386, opened 2017-09-02
Fix #350

### LINKED ISSUE #350: Logic operator's execute order is wrong
Example:

```ruby
a || b || c
```
In Ruby if `b` is not a false value, than the expression returns.
But in Goby it'll execute like `(a || (b || c))`. This is because we treat the operators as method instead of special keyword.

st0012:
We used to treat logic operators `&&` and `||` as methods. But that will cause unexpected expression evaluation.

For example:

```ruby
a = 0
true || a = 10
a
```

In our old implementation, `a` will be `10` because Goby evaluates expressions on `||`'s both sides. But actually it should stop when it has `true` as first value, so `a` should still be `0` because `a = 10` won't be evaluated.

For more example cases please see `vm/boolean_test.go`

This closes #350


hachi8833:
Makes sense and LGTM!

Maxwell-Alexius:
I think I've got it. LGTM

PR #390, opened 2017-09-06
Regexp class is coming (revised)

hachi8833:
Just to recreate the WIP PR to catch up the recent updates.


st0012:
Looking forward to merge this PR 👍

st0012:
@hachi8833 Do you have a timeline for this feature?

hachi8833:
Now I changed my mind that we should merge the current branch first. After that the following should be included (which need modification on parser):

- regex literal `//`
- match operator `=~`

vm/regexp_test.go
```diff
@@ -0,0 +1,61 @@
...
+	for i, tt := range tests {
+		v := initTestVM()
+		evaluated := v.testEval(t, tt.input, getFilename())
+		checkExpected(t, i, evaluated, tt.expected)
+		v.checkCFP(t, i, 0)
+		v.checkSP(t, i, 1)
+	}
+}
+
+func TestRegexpClassCreation(t *testing.T) {
```
st0012: Are you planning to uncomment those test cases? Otherwise I think this test function is unnecessary.

vm/classes/classes.go
```diff
@@ -15,4 +15,5 @@ const (
 	PluginClass   = "Plugin"
 	GoObjectClass = "GoObject"
 	FileClass     = "File"
+	RegexpClass   = "Regexp"
```
st0012: Why do you name it `Regexp` instead of `Regex`? What does the `p` mean?

vm/regexp_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	for i, tt := range tests {
+		vm := initTestVM()
+		evaluated := vm.testEval(t, tt.input, getFilename())
+		checkExpected(t, i, evaluated, tt.expected)
+		vm.checkCFP(t, i, 0)
+	}
+}
+
+//func TestRegexpMatch(t *testing.T) {
```
st0012: You should uncomment and add more test cases to this test function

st0012:
@hachi8833 Are you going to update and make this mergeable?

vm/classes/classes.go
saveriomiroddi: I think the `p` is part of `Exp`ression. Ruby names the regular expression class `Regexp` rather than `Regex`.

vm/regexp_test.go
hachi8833: I'm adding some test patterns.

saveriomiroddi:
@st0012 @hachi8833 if nobody's actively working on this, I could complete the task. I'm very interested in this, as I think it's a good intermediate step.

Regardless, I would remove the `.DS_Store` file, since it's MacOS proprietary stuff which gets in the way. Even better, I would remove it and add a rule to `.gitignore`.

st0012:
@saveriomiroddi You can just ask @hachi8833, I don't have any opinion on this. And you should just send a PR for update gitignore.

hachi8833:
@saveriomiroddi thank you for the offering! 
I would push the current branch with several additional tests soon so that you can join the implementations.
And I'd remove the unnecessary `.DS_Store` and check .gitignore

vm/regexp_test.go
hachi8833: Uncommented

hachi8833:
Sorry for the delay. Pushed the updated branch.

vm/regexp.go
```diff
@@ -0,0 +1,139 @@
...
+	Regexp *regexp2.Regexp
+}
+
+// Class methods --------------------------------------------------------
+func builtInRegexpClassMethods() []*BuiltinMethodObject {
+	return []*BuiltinMethodObject{
+		{
+			Name: "new",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
```
saveriomiroddi: Does this need a check for the number of arguments? (I'm not a 100% sure if the parser does this check or not)

vm/vm.go
```diff
@@ -201,8 +201,9 @@ func (vm *VM) initConstants() {
 		vm.initChannelClass(),
 		vm.initGoClass(),
 		vm.initFileClass(),
+		vm.initRegexpClass(),
```
saveriomiroddi: Commit f5e3e20b introduces, likely accidentally, a trailing space on this line, and a misaligned brace two lines after - I think they should be reverted.

hachi8833: Got it

vm/classes/classes.go
st0012: @saveriomiroddi I see, thanks for the explanation 😄

vm/regexp.go
st0012: I think this method should check argument numbers, and check it here.
@hachi8833 Can you make sure only one argument should be passed into this method?

vm/regexp_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{`Regexp.new("あい|いう|うえ").match?('いうう')`, true},
+		{`Regexp.new("(ととち)\1").match?('ととちととちととち')`, true},
+		{`Regexp.new("|え").match?('え')`, true},
+		{`Regexp.new("^あず").match?('あずあず')`, true},
+		{`Regexp.new("あず$").match?('あずあず')`, true},
+		{`Regexp.new("(((.あ)))\3").match?('zあzあああ')`, true},
+		{`Regexp.new("(あう*?ん)\1").match?('ああううんあううんあうん')`, true},
+		{`Regexp.new("ああん{3,4}").match?('ててああいいああんんんああんああん')`, true},
+		//{`Regexp.new("\000あ").match?("い\000あ")`, true},
+		//{`Regexp.new("とと\xfe\xfe").match?("ととと\xfe\xfe")`, true},
```
st0012: @hachi8833 What problem do we have when running these tests? Is caused by the lack of functionalities of Goby parser?

hachi8833: I guess this comes from dlclark's regexp package, not from Goby's parser.
I'll send the issue to him soon.

st0012:
@hachi8833 Is this ready to merge?

hachi8833:
Now it should be ready to merge. (updated the awkward code)

vm/regexp.go
st0012: @hachi8833 You should check the argument's number and its type in here, before you initialize and regexp object

hachi8833: Fixed

PR #391, opened 2017-09-07
Refactor: reorder functions in vm's classes

hachi8833:
Just to make the order of functions consistent, as well as changing the names as:

- builtIn -> builtin
- Builtin -> Builtin


hachi8833:
Rebasing was incorrect. Closing...

Issue #331, opened 2017-08-15
Support Form-Data Parsing to JSON in HTTP Request Object


Maxwell-Alexius:

saveriomiroddi:
@Maxwell-Alexius could you give some context/references/examples? I know nothing of the subject, and seeing something concrete would help me in knowing what to study.

PR #394, opened 2017-09-07
[WIP] Feature/http upgrades

ear7h:
Implemented an http client for fully customizable http requests and tests. Labeled as WIP because errors still need to be tested and http errors reinplemented


ear7h:
Next step is to add `HTTPError`, I tried to by simply adding a const in the http.go file

vm/http_client.go
```diff
@@ -0,0 +1,119 @@
...
+	uObj, ok := gobyReq.instanceVariableGet("@url")
+	if !ok {
+		return nil, errors.New("could not get url")
+	}
+
+	u := uObj.(*StringObject).value
+
+	methodObj, ok := gobyReq.instanceVariableGet("@method")
+	if !ok {
+		return nil, errors.New("could not get method")
```
st0012: You can just use `fmt.Errorf` to return an error object, so you can avoid using the builtin errors package.

vm/http.go
```diff
@@ -127,6 +135,26 @@ func builtinHTTPClassMethods() []*BuiltInMethodObject {
...
+		}, {
+			// Sends a HEAD request to the target with type header and body. Returns the HTTP response as a string.
+			Name: "head",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+					if len(args) != 1 {
+						return t.vm.initErrorObject(errors.ArgumentError, "Expect 1 arguments. got=%v", strconv.Itoa(len(args)))
+					}
+
+					host := args[0].(*StringObject).value
```
st0012: I think you might need to check the argument's type. Otherwise the Goby will crash if someone accidentally passed an integer variable or something like that.

vm/http.go
```diff
@@ -45,12 +47,22 @@ func initResponseClass(vm *VM, hc *RClass) *RClass {
 	hc.setClassConstant(responseClass)
 	builtinHTTPResponseInstanceMethods := []*BuiltInMethodObject{}
 
-	responseClass.setBuiltInMethods(builtinHTTPResponseInstanceMethods, false)
+	responseClass.setBuiltInMethods(builtinHTTPResponseInstanceMethods, true)
```
st0012: If the second argument is true then those methods will be set as class methods, why change this?
And if you want it to be class methods, please also change `builtinHTTPResponseInstanceMethods` to `builtinHTTPResponseClassMethods`

vm/http.go
```diff
@@ -45,12 +47,22 @@ func initResponseClass(vm *VM, hc *RClass) *RClass {
...
 
 	httpResponseClass = responseClass
 	return responseClass
 }
 
+func initClientClass(vm *VM, hc *RClass) *RClass {
+	clientClass := vm.initializeClass("Client", false)
+	hc.setClassConstant(clientClass)
+
+	clientClass.setBuiltInMethods(builtinHTTPClientClassMethods(), false)
```
st0012: Here has same issue like above.

ear7h:
Need to resolve merge conflicts

vm/http.go
```diff
@@ -24,24 +25,32 @@ func builtinHTTPClassMethods() []*BuiltinMethodObject {
 			Name: "get",
 			Fn: func(receiver Object) builtinMethodBody {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+					arg0, ok := args[0].(*StringObject)
+					if !ok {
+						return t.vm.initErrorObject(errors.ArgumentError, "Argument 0 must be a string")
```
st0012: You should write

```go
t.vm.initErrorObject(errors.ArgumentError, "First argument must be a String. got: %s", args[0].class.name)
```

vm/http.go
```diff
@@ -24,24 +25,32 @@ func builtinHTTPClassMethods() []*BuiltinMethodObject {
...
+					uri, err := url.Parse(arg0.value)
 
 					if len(args) > 1 {
 						var arr []string
 
 						for _, v := range args[1:] {
-							arr = append(arr, v.(*StringObject).value)
+							argn, ok := v.(*StringObject)
+							if !ok {
+								return t.vm.initErrorObject(errors.ArgumentError, "Splat arguments must be a string")
```
st0012: Here should be:

```go
t.vm.initErrorObject(errors.ArgumentError, "%d's splat argument must be a String. got: %s", i, args[i].class.name)
```

Which means we will need `_` to be `i`.

vm/http.go
```diff
@@ -65,17 +74,25 @@ func builtinHTTPClassMethods() []*BuiltinMethodObject {
 						return t.vm.initErrorObject(errors.ArgumentError, "Expect 3 arguments. got=%v", strconv.Itoa(len(args)))
 					}
 
-					uri, err := url.Parse(args[0].(*StringObject).value)
-					if err != nil {
-						return t.vm.initErrorObject(errors.ArgumentError, err.Error())
+					arg0, ok := args[0].(*StringObject)
+					if !ok {
+						return t.vm.initErrorObject(errors.ArgumentError, "Argument 0 must be a string")
```
st0012: Remember to change here's messages too

vm/http_client.go
```diff
@@ -0,0 +1,119 @@
...
+					result := t.builtinMethodYield(blockFrame, req)
+
+					if err, ok := result.Target.(*Error); ok {
+						fmt.Printf("Error: %s", err.Message)
+						return err //a Error object
+					}
+
+					goReq, err := requestGobyToGo(req)
+					if err != nil {
+						return t.vm.initErrorObject(gerrors.ArgumentError, "Request object incomplete object %s", err)
```
st0012: This message's grammar seems a little weird. @hachi8833 Can you take a took?

vm/http_client_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+		c = Net::HTTP::Client.new
+
+		res = c.send do |req|
+			req.url = "http://127.0.0.1:3001"
+			req.method = "GET"
+		end
+
+		res
+		`, "InternalError: Could not get response: Get http://127.0.0.1:3001: dial tcp 127.0.0.1:3001: getsockopt: connection refused", 6},
```
st0012: I thought you said you will add `HTTPError`?

st0012: And this message has duplicated informations. If we want to make users know what's going on in our way, we shouldn't still pass the message we get from Go.

st0012: And this message has duplicated informations. If we want to make users know what's going on in our way, we shouldn't still pass the message we get from Go.

vm/http_client.go
ear7h: I'll change it to "Request object incomplete %s", my mistake

ear7h:
Needs restructuring, as 'send' is a global method

vm/http_test.go
```diff
@@ -91,17 +91,17 @@ func TestHTTPObjectFail(t *testing.T) {
...
 
 		Net::HTTP.get("http://127.0.0.1:3000/error")
-		`, "InternalError: 404 Not Found", 4},
+		`, "HTTPError: Non-200 response, 404 Not Found (404)", 4},
 		{`
 		require "net/http"
 
 		Net::HTTP.post("http://127.0.0.1:3000/error", "text/plain", "Let me down")
-		`, "InternalError: 404 Not Found", 4},
+		`, "HTTPError: Non-200 response, 404 Not Found (404)", 4},
```
st0012: @ear7h I think we shouldn't raise error just because the response code is not 200. It's very normal to get response code other then 2xx. So it might be better to let user know the situation and handle that themselves.

ear7h: This class method is more more of a 'quick and dirty' approach which only returns the response body directly as a string. This behavior is consistent with ruby's `Net::HTTP.get` method (ruby does not implement a quick post method, only post form). 
When using the client with `Net::HTTP.start` errors are not thrown on non-200 responses.

st0012: Ok, but you need to clearly document this for others to know the result and reason.

ear7h:
The latest commit has comments in the class methods

st0012:
@ear7h It seems that this PR is almost completed, what do we still lack for now?

st0012:
@ear7h I think we still need some test cases for error situations to keep the test coverage. See [here](https://codecov.io/gh/goby-lang/goby/compare/e5e8015e51339e0efd961690a3c0311c6df801d4...b94986a645806c3032a7254b89073f75b2330433/src/vm/http.go)

ear7h:
@st0012 At the moment, the client class uses the default Go client with no way of changing it. I need to implement cookies and other settings

st0012:
@ear7h I think we can merge this first and open a new PR for the things you talk about. How do you think?

ear7h:
@st0012 I think that's good idea, I have no idea how i'm going to attack these changes

st0012:
@ear7h Sorry that I didn't ask until now, but does this PR also covers #331?

PR #398, opened 2017-09-09
Add Object#send.

### LINKED ISSUE #397: Implement send method

st0012:
This closes #397

### DIFF
--- a/vm/class_test.go
+++ b/vm/class_test.go
@@ -170,11 +170,27 @@ func TestCustomClassConstructor(t *testing.T) {
 	v.checkSP(t, 0, 1)
 }
 
-func TestDefClassMethod(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ear7h:
Should there be module tests as well?

vm/class_test.go
```diff
@@ -170,29 +170,61 @@ func TestCustomClassConstructor(t *testing.T) {
...
+		{`
+		class Foo
+		  def self.bar(x)
+		    10 + x
+		  end
+		end
+
+		Foo.send(:bar, 5)
+		`, 15},
 	}
```
ear7h: Should there be module tests as well?

st0012: Class and module are basic the same internally, so I don't think it's necessary

PR #407, opened 2017-09-11
[ci skip] Correct an error in Array#at() method documentation

saveriomiroddi:
The method documentation differs from the implementation (and the Ruby reference) - no error is raised in case of out-of-bound access, instead, nil is returned.


saveriomiroddi:
There is probably a better wording than the one I've used :-)

PR #408, opened 2017-09-11
Implement Array#delete_at(i)

saveriomiroddi:
Delete the element at the specified index.

PR #409, opened 2017-09-12
Make travis work on forked repositories

saveriomiroddi:
Goby's Travis setup won't work on forked repositories, as the cloned location is different from what the `import` statements expect.

Fixing the problem can't be accomplished by renaming the repository, since the goby project parent name (eg. `saveriomiroddi`) is the owner's/organization's name, and can't be changed.

The Travis `go_import_path` doesn't work as intended, so the logic is simply to move the directory to make it compatible with the `import`s.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
   - sudo apt-get install apache2-utils
 
 before_script:
+  - ./prepare_travis_for_fork.sh
   - psql -c 'create database goby_test;' -U postgres
 
 script:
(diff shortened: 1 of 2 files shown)


st0012:
@saveriomiroddi I'll merge this and can you quickly find someone to test it?

saveriomiroddi:
@st0012 if you can put me in touch with somebody from the organization, I can test with him/her.

For your reference, and verification, I've cherry picked this commit into #408, and you can see that now it's green (if/when you intend to merge #408, let me know, and I'll rebase, in order to keep a clean history).

PR #410, opened 2017-09-12
Add the packaging metadata to build the goby snap

elopio:
This package will let you publish the latest goby in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,27 @@
+name: goby
+version: master
+summary: A new language helps you develop microservices
+description: |
+  Goby is an object-oriented interpreter language deeply inspired by Ruby as
+  well as its core implementation by 100% pure Go. Moreover, it has standard
+  libraries to provide several features such as the Plugin system.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: devmode
+
+apps:
+  goby:
+    environment:
+      GOBY_ROOT: $SNAP
+    command: goby
+
+parts:
+  goby:
+    source: .
+    plugin: go
+    go-importpath: github.com/goby-lang/goby
+    install:
+      cp -R lib/ $SNAPCRAFT_PART_INSTALL
+    after: [go]
+  go:
+    source-tag: go1.9


elopio:
This requires more testing before bumping the grade to stable. I used goby to test our upcoming release of snapcraft, and checked some basic functions.

st0012:
@elopio Ok, let's try this out

st0012:
@elopio Well I have trouble setting up a build. I can't use either `goby` or `goby-lang` as the name, it says Forbidden. Do you know how can I fix this?

elopio:
@st0012 wooo \o/ Thanks for the merge.

That is probably a bug we have when the user doesn't yet have a developer name. Please take a look here, and let me know if that solves your problem:
https://github.com/canonical-websites/build.snapcraft.io/issues/945#issuecomment-327418956

I will be happy to help you with any doubt or problem that you have on your way to a stable release in the store. Just ping me, or post a topic in forum.snapcraft.io

pura vida.

st0012:
@elopio This solved the issue, thanks! But I think it might be better if I change the version in yaml file to our latest release, instead of just master, right?

elopio:
I'm sorry about that, we have had a long discussion about reducing the requirements for the store accounts. But you hitting the error gives us more information for that discussion, so thanks.

@st0012 if you hardcode the version, you will have to keep changing it. I left it `master`, because that's good enough to experiment. build.snapcraft.io leaves your snap in the edge channel, where it is only visible for people who know it's there, like your early adopters and the ubuntu community of testers after we make a call for testing. At this initial phase, the version value is not that important.

When you are ready for a wider audience, or even make it stable and release it to the store for everybody, it would be good to make the version update itself. We can use `version: git`, and snapcraft will be smart and use the name of the tag if it's releasing that commit. That only works for annotated git tags, which you are not using in goby. I would recommend you to start using them, it's a good practice: https://stackoverflow.com/questions/4971746/why-should-i-care-about-lightweight-vs-annotated-tags

If you don't want to use annotated tags, `version` also take an arbitrary script to run. With that you can get unannotated tags, something like `version: git describe --tags`.

But well, having said aaaall of that, the version value is for you and your users. The snap can take any value you want, and any workflow that makes you happy is good for us.

pura vida.

PR #412, opened 2017-09-13
Implement Array#values_at

saveriomiroddi:
Implementation with Integer arguments.


vm/array_test.go
```diff
@@ -1247,3 +1247,51 @@ func TestArrayUnshiftMethod(t *testing.T) {
 		v.checkSP(t, i, 1)
 	}
 }
+
+func TestArrayValuesAtMethod(t *testing.T) {
```
st0012: Can you also add an empty array case?

saveriomiroddi: :+1: Done.

PR #413, opened 2017-09-13
Implement Array#any?

saveriomiroddi:
This implementation requires a block.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -335,6 +335,10 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
 					}
 
 					if blockFrame != nil {
+						if len(arr.Elements) == 0 {
+							t.callFrameStack.pop()
+						}
+
 						for _, obj := range arr.Elements {
 							result := t.builtinMethodYield(blockFrame, obj)
 							if result.Target.(*BooleanObject).value {
@@ -463,6 +467,11 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
 
 					arr := receiver.(*ArrayObject)
 
+					// If it's an empty array, pop the block's call frame
+					if len(arr.Elements) == 0 {
+						t.callFrameStack.pop()
+					}
+
 					for _, obj := range arr.Elements {
 						t.builtinMethodYield(blockFrame, obj)
 					}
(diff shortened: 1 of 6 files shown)


vm/array_test.go
```diff
@@ -205,6 +205,77 @@ func TestArrayPlusOperatorFail(t *testing.T) {
...
+			  true
+			end
+		`, false},
+	}
+
+	for i, tt := range tests {
+		v := initTestVM()
+		evaluated := v.testEval(t, tt.input, getFilename())
+		checkExpected(t, i, evaluated, tt.expected)
+		v.checkCFP(t, i, 1)
```
saveriomiroddi: @st0012 I'm not sure about this. Originally I thought this was an issue (I don't how exactly what CFP is), then I found that `Array#select` yield the same result (in testing) if a test case with an empty block is performed.

vm/array.go
```diff
@@ -167,6 +167,59 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
...
+					if blockFrame == nil {
+						return t.vm.initErrorObject(errors.InternalError, errors.CantYieldWithoutBlockFormat)
+					}
+
+					for _, obj := range arr.Elements {
+						result := t.builtinMethodYield(blockFrame, obj)
+
+						booleanResult, isResultBoolean := result.Target.(*BooleanObject)
+
+     				if isResultBoolean {
```
st0012: Please fix these two line's indentation

vm/array_test.go
st0012: What do you mean yield the same result? You mean also executes the block?

st0012: And I know why the CFP is `1` instead of `0`.
The CFP stands for call frame pointer. In Goby we separate value stack and method call stack. And when we having a block argument, a block's call frame will be pushed on the call stack. And we pop that block's call frame after we execute (`yield`) the block.
But in this case you give it an empty array, which means the for loop won't be evaluated. So no yield action will be performed, the block's call frame just stays there.

st0012: BTW, thanks for spotting this issue, you can see how I fix this issue in [here](https://github.com/goby-lang/goby/commit/e3372b7306273e6c6e8937d69d8ee42e75932631)

vm/array.go
saveriomiroddi: Fixed. Sorry for that - my editor hides differences between spaces and tabs, so the code looks normal to me when committing. I actually have a proposal about how to make git handle this automatically.

saveriomiroddi: OK, ignore the proposal. I will try to find, for this project, some way to automate the conversion of spaces to tabs.

vm/array_test.go
saveriomiroddi: Thanks! Fixed, and updated the UT.

PR #414, opened 2017-09-14
Exclude incorrectly parsed files from codeclimate checks

saveriomiroddi:
Due to CC not currently supporting Go 1.9 syntax, some files are not parsed correctly, and cause CC to fail.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,8 +3,14 @@ engines:
     enabled: true
   golint:
     enabled: true
+    exclude_paths:
+      - "vm/instruction.go"
+      - "vm/vm.go"
   gofmt:
     enabled: true
+    exclude_paths:
+      - "vm/instruction.go"
+      - "vm/vm.go"
 
 exclude_paths:
   - "Godeps"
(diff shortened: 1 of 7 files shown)


saveriomiroddi:
@st0012 I'm not sure why CC fails on this PR. I've tested the change in my fork, and it yields green. 

Feel free to close this PR without merging if you think CC should still run on those two files :smile: 

I think it would be better to skip the checks on those 2 files (and revert the change when CC is fixed), since they don't add any value, and make the PR status red.

st0012:
I think I'll just wait for codeclimate to fix that issue, so I don't plan to change its config file. I'll check every PR's other CI status so it doesn't matter CC is red.

st0012:
@saveriomiroddi And I think you shouldn't include other PR's change (like this), in other PRs

saveriomiroddi:
> And I think you shouldn't include other PR's change (like this), in other PRs

Sorry, it was accidental - I tried it in master because CC was only testing my master branch, and not the other PRs.

I've now removed it from everywhere.

PR #418, opened 2017-09-15
[WIP] Add support for keyword arguments

shuheiktgw:
Still under implementation...

### DIFF
--- a/compiler/bytecode/statement_generation.go
+++ b/compiler/bytecode/statement_generation.go
@@ -11,6 +11,7 @@ const (
 	NormalArg int = iota
 	OptionedArg
 	SplatArg
+	KeywordArg
 )
 
 func (g *Generator) compileStatements(stmts []ast.Statement, scope *scope, table *localTable) {
@@ -155,25 +156,33 @@ func (g *Generator) compileDefStmt(is *InstructionSet, stmt *ast.DefStatement, s
 	newIS.isType = MethodDef
 
 	for i := 0; i < len(stmt.Parameters); i++ {
-		var argType int
 		switch exp := stmt.Parameters[i].(type) {
 		case *ast.Identifier:
-			argType = NormalArg
 			scope.localTable.setLCL(exp.Value, scope.localTable.depth)
+
+			newIS.argTypes = append(newIS.argTypes, NormalArg)
 		case *ast.AssignExpression:
-			argType = OptionedArg
 			exp.Optioned = 1
 			g.compileAssignExpression(newIS, exp, scope, scope.localTable)
+
+			newIS.argTypes = append(newIS.argTypes, OptionedArg)
 		case *ast.PrefixExpression:
 			if exp.Operator != "*" {
 				continue
 			}
-			argType = SplatArg
 			ident := exp.Right.(*ast.Identifier)
 			scope.localTable.setLCL(ident.Value, scope.localTable.depth)
-		}
 
-		newIS.argTypes = append(newIS.argTypes, argType)
+			newIS.argTypes = append(newIS.argTypes, SplatArg)
+		case *ast.HashExpression:
+			for key, exp := range exp.Data {
+				g.compileExpression(newIS, exp, scope, scope.localTable)
+				index, depth := scope.localTable.setLCL(key, scope.localTable.depth)
+				newIS.define(SetLocal, exp.Line(), depth, index, 1)
+
+				newIS.argTypes = append(newIS.argTypes, KeywordArg)
+			}
+		}
 	}
 
 	if len(stmt.BlockStatement.Statements) == 0 {
--- a/compiler/bytecode/statement_generation.go
+++ b/compiler/bytecode/statement_generation.go
@@ -11,7 +11,8 @@ const (
 	NormalArg int = iota
 	OptionedArg
 	SplatArg
-	KeywordArg
+	RequiredKeywordArg
+	OptionalKeywordArg
 )
 
 func (g *Generator) compileStatements(stmts []ast.Statement, scope *scope, table *localTable) {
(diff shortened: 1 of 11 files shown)


st0012:
With this feature we can port Ruby gems to Goby libraries 😄 
But I also need to implement basic library require mechanism first ~"~

st0012:
@shuheiktgw Just ping me if you need any help

vm/statement_test.go
```diff
@@ -156,6 +156,20 @@ func TestDefStatement(t *testing.T) {
...
+
+		foo(a:10)
+		`, 10},
+		{`
+		def foo(a:20)
+		  a
+		end
+
+		foo
+		`, 20},
```
shuheiktgw: @st0012 I'm kind of stuck here... The last test does not pass, and I believe I set default value at `line 178 - 192 in satement_generation.go`, but it raises out of index error probably at `line 281 in thread.go`. Could you help me out here?

Also, if you have any adviser for overall implementation, I'm glad to here that, too! (Although 40% of implementation has done so far I feel..)

shuheiktgw: Ok, I just noticed I should not call `t.stack.Data[argPr+argIndex]` when default value is not give! So, I kind of solved this problem, but still I'd appreciate it if would give me an advise! haha

shuheiktgw: BTW, I'm aware that my current implementation depends on the order of arguments, so `test(a: 1, b:2)` and `test(b: 2, a: 1)` give a different result. I'm going to work on the problem next.

vm/thread.go
```diff
@@ -204,12 +204,12 @@ func (t *thread) evalMethodObject(receiver Object, method *MethodObject, receive
...
 
 	for _, at := range method.argTypes() {
-		if at == bytecode.NormalArg {
+		if at == bytecode.NormalArg || at == bytecode.RequiredKeywordArg {
 			minimumArgNumber++
 		}
 	}
 
-	if argC > method.argc && method.lastArgType() != bytecode.SplatArg {
+	if argC > method.argc && !method.isSplatArgIncluded() {
```
st0012: Maybe we should start a new issue to write down a table for all the possible cases and checks, so we can have a bigger picture about the argument checking. And that can also be the future documentation as well

vm/thread.go
```diff
@@ -204,12 +204,12 @@ func (t *thread) evalMethodObject(receiver Object, method *MethodObject, receive
 	argTypesCount := len(method.argTypes())
 
 	for _, at := range method.argTypes() {
-		if at == bytecode.NormalArg {
+		if at == bytecode.NormalArg || at == bytecode.RequiredKeywordArg {
```
st0012: I think we should totally separate the keyword argument's check logic and other argument's check logic. Include required keyword argument into the minimum argument number count is not very intuitional for what/how we're going to check later.

compiler/bytecode/statement_generation.go
```diff
@@ -155,25 +157,40 @@ func (g *Generator) compileDefStmt(is *InstructionSet, stmt *ast.DefStatement, s
...
+			newIS.argTypes = append(newIS.argTypes, SplatArg)
+		case *ast.HashExpression:
+			for key, exp := range exp.Data {
+				var argType int
+
+				if exp == nil {
+					argType = RequiredKeywordArg
+				} else {
+					g.compileExpression(newIS, exp, scope, scope.localTable)
+					index, depth := scope.localTable.setLCL(key, scope.localTable.depth)
```
st0012: I think maybe we shouldn't use current local table here anymore. Because that only operate based on index, but what we need to do might need more information than that. So maybe we should upgrade our local table system, or use a new data structure for the keyword arguments.

compiler/parser/expression_parsing.go
```diff
@@ -163,6 +163,8 @@ func (p *Parser) parseIdentifier() ast.Expression {
 	return &ast.Identifier{BaseNode: &ast.BaseNode{Token: p.curToken}, Value: p.curToken.Literal}
 }
 
+
+
```
st0012: Remove these empty lines plz

compiler/bytecode/statement_generation.go
shuheiktgw: Yes, I agree that current index based operation already reached its limit. Let me think about how to upgrade local table so that somehow it can handle keyword arguments and clear up `evalMethodObject` method. (Also, I think we can improve argTypes to handle keyword argument, but I'm pretty sure it will end up with messing up `evalMethodObject` method even more....)

shuheiktgw: I'm referring to YARV's local table implementation now

st0012: Great job, but maybe we should upgrade the local table at another branch (base on current master). Because that'll be another new (and might be large) refactoring I think.

shuheiktgw: Hmm, do you think I should limit this pr to only parsing of keyword arguments? Since without refactoring local table class, I do not think it is possible to implement  code generation part of keyword argument.

st0012: Nope 
I mean you should implement local table in other PR, and rebase/merge master this once it's merged into master

shuheiktgw: You got it! Let me work on that on another pr! 👍

shuheiktgw: @st0012 Hmm, I'm still stuck in here... Actually I figured out that Ruby use a hidden argument when they call a method with a keyword argument. For example `test` method like below

```
def test(a, b: 1, c: 2); end
```
When Ruby call `test` method like this

```
test(10, b: 20, c: 30)
```

Ruby actually push the hash of defaults as a fourth argument, so now the arguments are those four below.
```
10
b: 20
c: 30
{ b: 1, c: 2 }
```

I tried to do the same thing but I could not figure out the way I could access the default hash from `thread.go`. I probably need to improve typed for that purpose but do you have any better ideas?

Thanks!

st0012: @shuheiktgw Ok, I'll also start working on this on the weekend.

shuheiktgw: You got it! Thanks!

shuheiktgw:
I will close this PR

PR #420, opened 2017-09-15
Merge `[]` and `at()` methods code

saveriomiroddi:
The implementation is a direct copy of the previous `[]` code.


saveriomiroddi:
The methods were already present before I started adding new ones :smile:

vm/array.go
```diff
@@ -990,6 +944,34 @@ func (a *ArrayObject) toJSON() string {
 	return out.String()
 }
 
+// Retrieves an object in an array using Integer index; common to `[]` and `at()`.
+func (a *ArrayObject) index(t *thread, args []Object, blockFrame *callFrame) Object {
```
st0012: Why passing blockFrame? Seems that it's not required for this method.

saveriomiroddi: Right! Removed.

PR #421, opened 2017-09-19
Add #methods

hanachin:
with @siman-man / @tompng at RubyKaigi YassLab house

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -830,7 +830,7 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 			Name: "methods",
 			Fn: func(receiver Object) builtinMethodBody {
 				return func(t *thread, args []Object, blockFrame *callFrame) Object {
-					methods := []interface{}{}
+					methods := []Object{}
 					set := map[string]interface{}{}
 					klasses := receiver.Class().ancestors()
 					if receiver.SingletonClass() != nil {
@@ -840,11 +840,11 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 						for _, name := range klass.Methods.names() {
 							if set[name] == nil {
 								set[name] = true
-								methods = append(methods, name)
+								methods = append(methods, t.vm.initStringObject(name))
 							}
 						}
 					}
-					return t.vm.initObjectFromGoType(methods)
+					return t.vm.initArrayObject(methods)
 				}
 			},
 		},
--- a/vm/class.go
+++ b/vm/class.go
@@ -826,6 +826,28 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 				}
 			},
 		},
+		{
+			Name: "methods",
+			Fn: func(receiver Object) builtinMethodBody {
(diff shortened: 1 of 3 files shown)


hanachin:
@st0012 I refactored the code, would you please review it?

vm/class.go
```diff
@@ -1076,6 +1098,19 @@ func (c *RClass) setAttrAccessor(args interface{}) {
 	c.setAttrWriter(args)
 }
 
+func (c *RClass) ancestors() []*RClass {
```
st0012: @hanachin Hice!

st0012: Looks like you can also implement the `ancestors` method later very easily?

vm/class.go
```diff
@@ -826,6 +826,28 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 				}
 			},
 		},
+		{
+			Name: "methods",
+			Fn: func(receiver Object) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *callFrame) Object {
+					methods := []interface{}{}
```
st0012: Can't we just use `[]string{}` here? Seems like we'll only append strings into this slice.

hanachin: `initObjectFromGoType` only accepts `[]interface{}{}`, that's why I use `[]interface{}[]`.

When I change the `[]interface{}{}` to `[]string{}`, I got following error.

```
/private/var/folders/fp/3_xdx8zs3mb_mmtkqsxz8m1w0000gn/T/___TestMethods_in_class_test_go -test.v -test.run ^TestMethods$
	vm_test.go:281: At test case 0: ERROR: UndefinedMethodError: Undefined Method 'first' for <GoObject: 0xc42013f200>. At /Users/sei/src/github.com/goby-lang/goby/vm/class_test.go:9
	vm_test.go:182: At case 0 expect main thread's cfp to be 0. got: 1
	vm_test.go:281: At test case 1: ERROR: UndefinedMethodError: Undefined Method 'first' for <GoObject: 0xc42018c080>. At /Users/sei/src/github.com/goby-lang/goby/vm/class_test.go:10
	vm_test.go:182: At case 1 expect main thread's cfp to be 0. got: 1
	vm_test.go:281: At test case 2: ERROR: UndefinedMethodError: Undefined Method 'first' for <GoObject: 0xc4201a6ec0>. At /Users/sei/src/github.com/goby-lang/goby/vm/class_test.go:9
	vm_test.go:182: At case 2 expect main thread's cfp to be 0. got: 1
	vm_test.go:281: At test case 3: ERROR: UndefinedMethodError: Undefined Method 'include?' for <GoObject: 0xc4201c3800>. At /Users/sei/src/github.com/goby-lang/goby/vm/class_test.go:4
	vm_test.go:182: At case 3 expect main thread's cfp to be 0. got: 1
```

vm/class.go
hanachin: Yes! 🍻

vm/class.go
hanachin: That's right! I updated the code.

st0012:
@hanachin Nice work, thanks!

hanachin:
Thank you for the review 😄

Issue #427, opened 2017-09-27
Add #each method to Hash class


st0012:
Currently we only have `#each_key` and `#each_value`, but don't have `#each` method.

saveriomiroddi:
Addressed by #429.

PR #429, opened 2017-09-28
Implement Hash#each

saveriomiroddi:
In order to facilitate testing, the testing API `testBidimensionalArrayObject` has been added.

PR #431, opened 2017-09-29
Add each method to HashObject

janczer:
I added `each` method to Hash. (#427)

```
{a: 1, b: 2}.each do |k, v|
    puts(k + " " + v.to_s)
end
# => a 1
# => b 2
```

### DIFF
--- a/vm/hash.go
+++ b/vm/hash.go
@@ -254,6 +254,50 @@ func builtinHashInstanceMethods() []*BuiltinMethodObject {
 			},
 		},
 		{
+			// Loop through key and values of the hash with given block frame. It also returns array of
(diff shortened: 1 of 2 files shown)


saveriomiroddi:
Duplicate of #429.

janczer:
@saveriomiroddi Unfortunately I did't see it before.

saveriomiroddi:
@janczer :smile: If you are planning to work on Hash or Array methods, we can arrange and split the work, and I'm going through those classes, and we risk to duplicate the work again!

If you intend to do so, let me know which methods you'll work on, and I'll work on something else :smile:

PR #433, opened 2017-09-29
Implement Hash#any?

saveriomiroddi:


vm/hash_test.go
```diff
@@ -187,6 +187,62 @@ func TestHashComparisonOperation(t *testing.T) {
...
 }
 
+func TestHashAnyMethod(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		{`
+      { a: 1, b: 2 }.any? do |k, v|
+        v == 2
```
st0012: So should it be `true` if we put a `5` here?

saveriomiroddi: Yes, the second test:

    a = { a: 1, b: 2 }
    
    a.any? do |k, v|
      v
    end            # => true

tests/documents this specification.

PR #435, opened 2017-09-30
Implement Hash#keep_if

saveriomiroddi:


saveriomiroddi:
@st0012 I've realized that if you intend to have destructive methods on Hash/Array (eg. `select!`), this method is actually redundant (which is not good for Goby design), so it can be closed (or better, I can rename the method to `select!`).

st0012:
Actually I don’t want any method that does mutation. So this method is fine for me.
I didn’t merge it because there were too many PRs and I usually updated some PRs but forgot to merge them, then I need to sync them again 😅

saveriomiroddi:
> Actually I don’t want any method that does mutation. So this method is fine for me.
I didn’t merge it because there were too many PRs and I usually updated some PRs but forgot to merge them, then I need to sync them again :sweat_smile:

Good to know. I think I forgot about the Goby design related to the destructive methods :laughing:

st0012:
@saveriomiroddi This looks good to me, but I don't want to mutate the original hash (I didn't actually know this until I took a closer look today, sorry), which will make this method act just like `#select`. So I'm not sure about adding this method. Any thoughts? @goby-lang/maintainers

st0012:
@saveriomiroddi And I know I already merged some methods like `delete_if`, which is also my fault (sorry again).
I think we should use this chance to discuss that if my thoughts on the mutation method is reasonable.

saveriomiroddi:
> This looks good to me, but I don't want to mutate the original hash (I didn't actually know this until I took a closer look today, sorry), which will make this method act just like #select.

No problem. This is a very interesting subject. I think use cases should be thought out carefully, to see which design can satisfy the typical uses cases.

Is there any sprcific reason not to use `Hash#new()` (which right now is disabled)?

I could see that satisfy both the immutability requirement, and the common use cases.

When I refer to common use cases, in particular, I refer to two:

1. `Hash.new(0)`: this is useful for hashes holding counters
2. `Hash.new { |h, k| h[k] = [] }`: this is useful for hash holding collections

saveriomiroddi:
> And I know I already merged some methods like delete_if, which is also my fault (sorry again).

From my perspective, immutability as design can be a reasonable point (even if I personally use mutability at work), as for collections, there are always duplicate mutable/immutable methods, and I don't think there are cases where not having mutable methods cause trickiness in the code.

st0012:
@saveriomiroddi My thoughts about Goby's design is very simple: reduce the chance for developer to screw things up. (especially for not very experienced developers).

Ruby gives a lot of powers to developers. This can let experienced developers get their work done more easily, but can also make it easier for those not experienced developers to screw things up. And I believe in reality the later case is more frequently seen than the first case.

So in Goby I want to change this. And I believe reducing mutation is one of the approaches, because unintentionally modifying data is often the causes of bugs.

(Maybe I should write done my view or expectation on Goby in somewhere?)

st0012:
BTW I'm not saying I don't like mutation at all (I'm not a big fan of those zero mutation, functional programming concept). I just think methods with side effects bring more troubles than advantages overall.

saveriomiroddi:
> (Maybe I should write done my view or expectation on Goby in somewhere?)

I think it's a good idea to have the design philosophy and/or an at least basic specification, so that people can have from the beginning a better idea about how to implement contributions (eg. what *not* to do :smile: ).

By the way, I've just realized that my discussion about `Hash.new` was in the wrong place, so ignore it :smile:

saveriomiroddi:
@st0012 based on the discussion above, I'm closing this PR.

PR #438, opened 2017-10-01
Implement the `:dig` method for `Hash` and `Array`

saveriomiroddi:
Implemented the `:dig` method for `Hash` and `Array`, via `Diggable` Go interface.

The PR contains also a small refactoring in the `Array` class, adding the internal `normalizedIndex()` method.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -412,6 +412,32 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
 			},
 		},
 		{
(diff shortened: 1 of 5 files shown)


saveriomiroddi:
@st0012 this implementation uses a Go interface.

There is an alternative, which is to implement `#dig` entirely in Goby. I'm neutral, if you don't like the current implementation, I can easily switch.

st0012:
@saveriomiroddi Nice work and refactoring 😄

PR #439, opened 2017-10-02
Implement Hash#fetch

saveriomiroddi:


vm/hash_test.go
```diff
@@ -907,49 +946,49 @@ func TestHashMapValuesMethod(t *testing.T) {
 		{`
 		h = { a: 1, b: 2, c: 3 }
 		result = h.map_values do |v|
-		  v * 3
```
st0012: @saveriomiroddi I prefer keeping the two spaces indentation for those Goby code in test cases

vm/hash_test.go
```diff
@@ -623,47 +746,17 @@ func TestHashEqualMethodFail(t *testing.T) {
...
-		`, nil},
-		{`
-		h = { a: 1, b: "Hello", c: true }.delete("a")
-		h["b"]
-		`, "Hello"},
-		{`
-		h = { a: 1, b: "Hello", c: true }.delete("a")
-		h["c"]
-		`, true},
+			{ spaghetti: "eat" }.fetch("pizza", "not eat")
```
st0012: Why don't you test the successfully fetching condition?

saveriomiroddi: You're right, that was an omission :smile:

saveriomiroddi:
@st0012 changes applied; I've entirely removed the commit with the indentation change.

Issue #379, opened 2017-08-31
Implement Primitive Data Type - Float


Maxwell-Alexius:

ear7h:
How are we doing on this? I could help implement this on the vm level, implementing builtin methods like [ruby](https://ruby-doc.org/core-2.2.0/Float.html). However, compiler/interpreter level implementations are a bit out of my reach

st0012:
@ear7h Because I have a talk next week (it's about Goby!) and I'm busy working on it, so this issue will be pending until I get back from Japan.

ear7h:
Alright, let me know if I can be of any assistance. Good luck btw!

saveriomiroddi:
Being addresses by #443.

st0012:
I will do the front-end (parsing) part if no one is interested in it. @Maxwell-Alexius Do you want to give it a try?

Maxwell-Alexius:
I think I can handle this part, let me try it. I will ping you in Slack

st0012:
I think we already implemented the Float data type, just need to improve it now. So closing this

PR #443, opened 2017-10-03
Float class (backend)

saveriomiroddi:
Backend implementation of the Float class, without the frontend (parsing). At this stage, Floats are instantiated via `String#to_f`.

Two related additions are made to the codebase:

- the `Numeric` interface, common to `Float` and `Integer`; currently it's only a lightweight, convenient, interface, not a full/visible Goby class
- the static method `toBooleanObject()`, which replaces the boilerplate for converting Go boolean values to a Goby boolean Object

This PR closes the backend side of #379.

### DIFF
--- a/vm/integer_test.go
+++ b/vm/integer_test.go
@@ -61,10 +61,10 @@ func TestIntegerArithmeticOperation(t *testing.T) {
 
 func TestIntegerArithmeticOperationFail(t *testing.T) {
 	testsFail := []errorTestCase{
-		{`1 + "p"`, "TypeError: Expect argument to be Integer. got: String", 1},
-		{`1 - "m"`, "TypeError: Expect argument to be Integer. got: String", 1},
-		{`1 ** "p"`, "TypeError: Expect argument to be Integer. got: String", 1},
-		{`1 / "t"`, "TypeError: Expect argument to be Integer. got: String", 1},
+		{`1 + "p"`, "TypeError: Expect argument to be Numeric. got: String", 1},
+		{`1 - "m"`, "TypeError: Expect argument to be Numeric. got: String", 1},
+		{`1 ** "p"`, "TypeError: Expect argument to be Numeric. got: String", 1},
+		{`1 / "t"`, "TypeError: Expect argument to be Numeric. got: String", 1},
 	}
 
 	for i, tt := range testsFail {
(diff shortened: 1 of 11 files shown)


saveriomiroddi:
Reference: Ruby strategy for cross-data type operations:

    /*
     * call-seq:
     *   float + other  ->  float
     *
     * Returns a new Float which is the sum of +float+ and +other+.
     */

    static VALUE
    flo_plus(VALUE x, VALUE y)
    {
        if (RB_TYPE_P(y, T_FIXNUM)) {
      return DBL2NUM(RFLOAT_VALUE(x) + (double)FIX2LONG(y));
        }
        else if (RB_TYPE_P(y, T_BIGNUM)) {
      return DBL2NUM(RFLOAT_VALUE(x) + rb_big2dbl(y));
        }
        else if (RB_TYPE_P(y, T_FLOAT)) {
      return DBL2NUM(RFLOAT_VALUE(x) + RFLOAT_VALUE(y));
        }
        else {
      return rb_num_coerce_bin(x, y, '+');
        }
    }

And (see https://silverhammermba.github.io/emberb/c/):

    There are a host of “NUM” macros that try to be more duck-typish about things. These will convert their C types to whatever Ruby Numeric subclass seems appropriate:
    
    ...
    DBL2NUM() for double

saveriomiroddi:
@st0012 should the equivalent of `Integer#to_intNN` be implemented for Float? For example, `Float#to_float32`? I don't know this design/concept, since there's nothing similar in Ruby.

st0012:
@saveriomiroddi The purpose for those methods is for calling Go functions. For current Integer class, those methods just change integer object's flag, and those flags only will be used when the integer is passed as Go function's parameters.

But I think I'll change the implementation of this by letting methods like `to_int8` return a `GoObject` that really contains the `int8` type integer. Because mutation is bad.

saveriomiroddi:
@st0012 ready for review.

vm/integer_test.go
```diff
@@ -22,32 +22,39 @@ func TestIntegerClassSuperclass(t *testing.T) {
...
-		{`25 / 5`, 5},
-		{`25 > 5`, true},
-		{`4 > 6`, false},
-		{`-5 < -4`, true},
-		{`100 < 81`, false},
-		{`5 ** 4`, 625},
-		{`25 / 5`, 5},
-		{`1 / 1 + 1`, 2},
-		{`0 / (1 + 1000)`, 0},
-		{`5 ** (3 * 2) + 21`, 15646},
```
st0012: @saveriomiroddi I want to keep these combined arithmetic operations test cases. Can you create another test function for these kind of tests?

st0012:
BTW thanks for reorganizing our integer tests at the same time 👍

vm/integer_test.go
saveriomiroddi: :+1: Restored!

st0012: Nice 😎

PR #445, opened 2017-10-05
Add support for case expression

shuheiktgw:
Added support for case expression! Thank you! 👍


compiler/parser/statement_parsing.go
```diff
@@ -258,7 +258,7 @@ func (p *Parser) parseBlockStatement() *ast.BlockStatement {
 		p.nextToken()
 	}
 
-	for !p.curTokenIs(token.End) && !p.curTokenIs(token.Else) && !p.curTokenIs(token.ElsIf) {
+	for !p.curTokenIs(token.End) && !p.curTokenIs(token.Else) && !p.curTokenIs(token.ElsIf) && !p.curTokenIs(token.When) {
```
shuheiktgw: I know this part is not good since block parsing stops whenever parser encounters those tokens. I'm thinking about giving tokens as arguments, which tell where to stop parsing block statement. I'll work on it in next PR.

shuheiktgw:
I've finished implementing case expression, so could anyone review this pr?

compiler/parser/expression_parsing.go
```diff
@@ -507,6 +507,92 @@ func (p *Parser) parseIfExpression() ast.Expression {
 	return ie
 }
 
+// Case expression forms if statement when parsing it
+//
+// ex.
+// case 1
```
st0012: Can you wrap the code like we normally do in the mark down?

compiler/parser/statement_parsing.go
st0012: I think you can first open a PR to refactor this part's logic if you could.

compiler/token/token.go
```diff
@@ -67,6 +67,9 @@ const (
 	If     = "IF"
 	ElsIf  = "ELSIF"
 	Else   = "ELSE"
+	Case   = "CASE"
+	When   = "WHEN"
+	Then   = "THEN"
```
st0012: @shuheiktgw When will we have to use the `then` keyword? I think it's redundant since most Rubyist ignore this keyword. Any other thoughts? @goby-lang/maintainers

st0012:
@goby-lang/maintainers This is kind of a big feature, can you all take a look on it? Also, this PR is great for learning how to add keywords and new statements in Goby.

compiler/token/token.go
shuheiktgw: I thought I think about porting ruby gems into Goby ecosystem, and if so, I think we need this... Otherwise I agree `then` keyword is redundant!  👍

st0012: @shuheiktgw Um....I think we should remove `then` when porting those gems instead of supporting it. And I don't see many people use this keyword nowadays actually, do you see a lot?

vm/evaluation_test.go
```diff
@@ -1637,6 +1637,113 @@ func TestIfExpressionEvaluation(t *testing.T) {
...
+			case 2
+			when 0
+			  0
+			when 1
+			  1
+			when 2
+			  2
+			end
+			`,
+			2,
```
st0012: I'll merge this first, but I hope you can add more test cases for this.

shuheiktgw:
Sorry for my late reply..., I will work on refactoring paseBlock and adding test cases for case statement!!

PR #447, opened 2017-10-05
Keyword argument support part 2

st0012:
This PR finishes the keyword argument's compilation.

And the most important change is the way we passing parameters/arguments information between compiler and VM. We use a new data structure called `ArgSet`. It records parameters/arguments name and their type (argument type) in order.

This is a little like a cheating, because this couples the compiler and the VM very heavily. Also, we can't properly do integration tests on compiler, because it's hard to print out this data structure.

But all these things will be fixed in the future PRs, at least we have keyword argument support now 😄


st0012:
Almost finished, please help me review this @goby-lang/maintainers

Issue #103, opened 2017-05-19
Range of Integer class


hachi8833:
I manually checked the edge cases of Integer class with the latest master:

```ruby
a =  999999999999999999             # works
a =  9999999999999999999            # => panic: could not parse "9999999999999999999" as integer

# upper limit
a =  9223372036854775807            # equivalent to 0x7FFFFFFFFFFFFFFF  (in 64-1 bit length)

# overrun
a =  9223372036854775807 + 1        # => -9223372036854775808 (negative number)
a =  9223372036854775808            # => panic: could not parse "9223372036854775808" as integer

# lower limit
a = -9223372036854775807            # equivalent to f8000000000000001  (in 64 bit length including sign)

# overrun
a = -9223372036854775807 - 1        # => -9223372036854775808
a = -9223372036854775807 - 2        # =>  9223372036854775807
a = -9223372036854775808            # => panic: could not parse "9223372036854775808" as integer
```

I don't think this is a bug but comes from Go's spec, as well as a specification matter of Goby. It might be good to add above to Integer's API doc.

------------------
Future: It might help us: https://golang.org/pkg/math/big/

Ruby 2.4 finally ceased the legacy FixNum and BigNum classes and migrated them to Integer, so I think switching between Integer and BigNum again in Goby might not be good.

st0012:
Thanks for reporting! I'll open a page to record some edge cases like this

hachi8833:
Fixed a typo and miscalculation in the report :-p

st0012:
@hachi8833 Please PM me on gitter when you're available if you're still finding things to contribute.

PR #449, opened 2017-10-05
Fix integer parsing panic

### LINKED ISSUE #448: Goby panics when parsing a too big Integer literal
Instead of panic, a parsing error should be returned.

Example:

    » 9223372036854775808
    panic: could not parse "9223372036854775808" as integer

    goroutine 1 [running]:
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseIntegerLiteral(0xc4201162d0, 0xb30df4, 0xc4201785f0)

saveriomiroddi:
Return an error instead of panic, when an Integer parsing error occurs.

Such error can happen under regular (non-programmtic error) circumstances, when the user inputs a too big integer.

Fixes #448.


st0012:
This closes #103 as well (actually #448 and #103 address the same issue), thanks for the fix!

PR #454, opened 2017-10-09
WIP: Full boolean values support, via new PutBoolean bytecode instruction

### LINKED ISSUE #450: Document (or fix, if it's a bug) the boolean string literals
I've noticed that in Goby, the string literals `"true"` and `"false"` directly translate to boolean values (respectively, true and false).

Since this behavior is unique among the scripting languages (Ruby/Python/Perl/PHP), I was wondering if this is intentional or not.

If intentional, I think this should be very carefully evaluated, in particular because of undefined/unexpected behaviors:

    if false; echo "not printed"; end #=> evaluates to false
    if "false"; echo "not printed"; end #=> evaluates to false
    if "False"; echo "printed!"; end #=> evaluates to true
    if "fa" + "lse"; echo "printed!"; end #=> evaluates to true (!)
    ("false").class #=> Boolean
    ("fa" + "lse").class #> String (!)

In essence, the strings `"false"`/`"true"` evaluate to Boolean, but not consistently - they do if they're literals, they don't if they're string variables.

If it's a bug, I can easily fix it (it's in `VM#initObjectFromGoType`).

saveriomiroddi:
Add the PutBoolean instruction, which completes the support for boolean values.

String values will now always evaluate to true in expressions:

    if "false"; puts "Executed"; end => "Executed"

Closes #450.


st0012:
Looks like just need some tests then this PR can be merged 👍

saveriomiroddi:
Yes, it's WIP :wink: I will likely complete it today! :+1:

saveriomiroddi:
@st0012 ok, I've added functional UTs; I've checked the internals, but currently, there are no test suites at all for them.

st0012:
You’re right, we currently don’t have any unit tests for VM. This is because we still keep adding features and change internal structures. So I think integration tests are more useful for now.
But if you want to add any unit tests I’m ok for it.

saveriomiroddi:
> You’re right, we currently don’t have any unit tests for VM. This is because we still keep adding features and change internal structures. So I think integration tests are more useful for now.
> But if you want to add any unit tests I’m ok for it.

I'm ok with the PR as it is, then :smile: 

For some reason, Hound is stuck on this PR, saying "Hound is busy reviewing changes...".

st0012:
@saveriomiroddi Sometimes they just stuck there, I guess it's because open source projects have low priority on the resource usage, since we're using it for free.

But normally I would check the coding style my self and I also lint/reformat master after some major PR s get merged.

PR #458, opened 2017-10-12
Refactor parse block to give tokens telling where to stop

shuheiktgw:
I refactored parseBlockStatement to give end tokens as an argument.  

For example, ` if statement` has to stop parsing block when the parser encounters `elsif`, `else`, `end`. On the other hand, `def statement` has to stop parsing block statement only if the parser encounters `end` like so.

Thank you!

### DIFF
--- a/compiler/parser/statement_parsing.go
+++ b/compiler/parser/statement_parsing.go
@@ -286,11 +286,11 @@ func (p *Parser) parseBlockStatement(endTokens ...token.Type) *ast.BlockStatemen
 		p.nextToken()
 	}
 
-OuterLoop:
+ParseBlockLoop:
 	for {
 		for _, t := range endTokens {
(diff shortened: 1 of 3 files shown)


compiler/parser/statement_parsing.go
```diff
@@ -286,7 +286,13 @@ func (p *Parser) parseBlockStatement() *ast.BlockStatement {
 		p.nextToken()
 	}
 
-	for !p.curTokenIs(token.End) && !p.curTokenIs(token.Else) && !p.curTokenIs(token.ElsIf) && !p.curTokenIs(token.When) {
+ParseBlockLoop:
```
st0012: @shuheiktgw I didn't know we can do this in before! Nice work 👍

st0012:
@shuheiktgw Nice refactoring, thanks!

PR #459, opened 2017-10-12
Implement MatchData

saveriomiroddi:
The `MatchData` class allows captures matching with regular expressions:

    md = Regexp.new("a (\w+)" ).match("This is a flower") #=> #<MatchData "a flower" 1:"flower">
    md.captures[0] #=> "flower"


hachi8833:
Amazing!
I think you can move `match` to Regexp. How about this?

saveriomiroddi:
> I think you can move match to Regexp. How about this?

@hachi8833 sorry, can you point it in the code (eg. via review)? if you refer to the method, it's already defined as `Regexp#match`.

@st0012 the comment above actually made me think about where regex matching methods should be located. in ruby, they're all located in `String`, so that when one wants to perform a match, the string is always the left element, and the regex the right one:

    string (operation) regex

right now, methods are begin added to Regexp (even this PR), so that there is an inconsistency between the invocation patterns:

    string =~ regex
    regexp.match?(string)

is this intended? do you have any design direction?

st0012:
@saveriomiroddi Putting the method in String class seems to be a better choice.

saveriomiroddi:
PR completed.

vm/match_data_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"testing"
+)
+
+func TestMatchDataClassSuperclass(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected string
+	}{
+		{`String.class.name`, "Class"},
```
st0012: I think this should be `MatchData`?

saveriomiroddi: Nuts! Fixed :smile:

PR #462, opened 2017-10-13
Refactor thread

st0012:
Currently `vm/thread.go`'s code is quite ugly, especially the method call part. So this PR is aim at refactoring thread's `evalMethodObject` function.


st0012:
@shuheiktgw Can you help me review this part? You're the only one who have touched it. I'm trying to extract method call logic from `evalMethodObject` function.

PR #464, opened 2017-10-16
Docker URL is 500 error in README.md

koduki:
Current docker URL is 500 error. Maybe correct URL is `https://hub.docker.com/r/gobylang/goby/` .


st0012:
@koduki You're right, thanks!

PR #470, opened 2017-10-20
Support Formal Front-End of Float Type Data

Maxwell-Alexius:
Currently using the Go `float64` datatype to support for Goby's float type data

Small issue to address - cases below will show the correct result

```ruby
(3.14159265358979).to_d.to_s # => "3.14159265358979"
(-273.150000000).to_d.to_s # => "-273.15"

# and without the round bracket
3.14159265358979.to_d.to_s # => "3.14159265358979"
```

However, there is a case when a negative valued float type number cannot call method without round bracket which I've labeled with a TODO
```ruby
-273.150000000.to_d.to_s # ERROR
```
See #526  

Another thing to address is that, currently this PR *only* support the basic format of Float API (and because this PR is big enough), so to implement the special format of float, e.g. the exponential format as described in issue #471, welcome to contribute another PR


Maxwell-Alexius:
@st0012  Can you hint me on writing tests about `parseFloatLiteral` or other kind of tests? I'm still a little confused and not sure about I'm on the right track. QAQ

vm/float_test.go
```diff
@@ -200,8 +200,8 @@ func TestFloatConversions(t *testing.T) {
 		input    string
 		expected interface{}
 	}{
-		{`'100.3'.to_f.to_i`, 100},
-		{`'100.3'.to_f.to_s`, "100.3"},
+		{`(100.3).to_i`, 100},
```
st0012: We should also support (or raise parsing error) when there's no parentheses around the float.

st0012:
@Maxwell-Alexius We'll have an large reconstruction for our compiler's tests, so don't worry about adding parser tests now.

saveriomiroddi:
For reference, programming languages generally support also the exponential format:

    2E2
    2e2
    20.0e-1

I think this should be either implemented in this PR, or a separate issue should be created, for tracking purposes.

saveriomiroddi:
For reference, programming languages generally support also the exponential format:

    2e2
    2E2
    2e-2
    20.0e3

I think this should be either addresses in this PR, or added as an issue for tracking purposes.

st0012:
@saveriomiroddi Thanks for reminding, I'll create another issue for this format.

Maxwell-Alexius:
@st0012 Currently there is a small problem -- cases below will show the correct result

```ruby
(3.14159265358979).to_d.to_s # => "3.14159265358979"
(-273.150000000).to_d.to_s # => "-273.15"

# and without the round bracket
3.14159265358979.to_d.to_s # => "3.14159265358979"
```

However, there is a case when a negative valued float type number cannot call method without round bracket which I've labeled with a TODO
```ruby
-273.150000000.to_d.to_s # ERROR
```

vm/float_test.go
```diff
@@ -200,15 +200,42 @@ func TestFloatConversions(t *testing.T) {
...
+		(-273.150000000).to_d.to_s`,
 			"-273.15"},
+		{`100.3.to_i`, 100},
+		{`100.3.to_s`, "100.3"},
+		{`100.3.to_d.to_s`, "100.3"},
+		{`
+		3.14159265358979.to_d.to_s`,
+			"3.14159265358979"},
+		// TODO: Able to parse negative float value and call method without parentheses
+		//{`
```
Maxwell-Alexius: Here is the small problem which is described in PR

Maxwell-Alexius:
Another thing to address is that, currently this PR *only* support the basic format of Float API (and because this PR is big enough), so to implement the special format of float, e.g. the exponential format as described in issue #471, welcome to contribute another PR

st0012:
@saveriomiroddi Do you want to take a look before I merge this?

compiler/parser/expression_parsing.go
```diff
@@ -388,6 +388,18 @@ func (p *Parser) parseMultiVariables(left ast.Expression) ast.Expression {
 	return result
 }
 
+func (p *Parser) parseDotExpression(receiver ast.Expression) ast.Expression {
+	_, ok := receiver.(*ast.IntegerLiteral)
+
+	// When both receiver & caller are integer => Float
+	if ok && p.peekTokenIs(token.Int) {
```
saveriomiroddi: Based on this code, expresions with spaces on the sides of the dot will be parsed as floats (eg. `2  .  3`).

If this is intended, I think it should be documented with a test case.

vm/float.go
```diff
@@ -438,7 +438,7 @@ func (f *FloatObject) numericComparison(t *thread, rightObject Object, operation
 // toString returns the object's value as the string format, in non
 // exponential format (straight number, without exponent `E<exp>`).
 func (f *FloatObject) toString() string {
-	return strconv.FormatFloat(f.value, 'f', -1, 64)
+	return strconv.FormatFloat(f.value, 'f', -1, 64) // fmt.Sprintf("%f", f.value)
```
saveriomiroddi: Is the comment intended?

compiler/parser/expression_parsing.go
st0012: Actually currently Goby doesn't recognize white space, so it's not just float has this issue, like:

```ruby
[] . length #=> 0
```

But this shouldn't be too hard to solve I think, already have a rough idea. But that won't be included in this PR.

PR #473, opened 2017-10-22
ConcurrentHash implementation

### LINKED ISSUE #469: Hash accessing is not thread-safe
Conclusion: We shouldn't use Go's `sync.Map` to implement normal hash, because it's performance it's poor in normal condition. [Reference](https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c)

saveriomiroddi:
Implementation of thread-safe associative arrays (Hash).

The implementation internally uses Go's `sync.Map` type, with some advantages and disadvantages:

- it is highly performant and predictable for a certain pattern of usage (`concurrent loops with keys that are stable over time, and either few steady-state stores, or stores localized to one goroutine per key.`); performance and predictability in other conditions are unspecified;
- iterations are non-deterministic; during iterations, keys may not be included;
- size can't be retrieved;
- for the reasons above, the Hash APIs implemented are minimal.

For details, see https://golang.org/pkg/sync/#Map.

Concurrent hashes are instantiated via `new()`:

    ConcurrentHash.new()
    ConcurrentHash.new({"a": 1, "b": 2})

Closes #469.


saveriomiroddi:
@st0012 the class currently misses the fuzzy testing, I'll add over the next days.

saveriomiroddi:
@st0012 what's the easiest way to directly test a method implementation? I'd like to do something similar to this:

		vm := initTestVM()
		regularHash := vm.initHashObject(map[string]Object{})
		regularHashGetter := regularHash.findMethod("[]").(*BuiltinMethodObject)
		result := regularHashGetter.Fn(regularHash)(vm.mainThread, []Object{NULL}, nil)

which fails at the last line, I'm not sure if it's because of the missing callFrame (I don't know how to prepare one, either).

st0012:
@saveriomiroddi For the last question, I suggest not to test things that way for now. Because currently the method's execution heavily depends on the VM's state, which is built with a series of instructions. So there's no a easy way to produce the environment. Also, produce Goby objects in Go is quite verbose if you want to write more complex tests.

vm/concurrent_hash.go
```diff
@@ -0,0 +1,168 @@
...
+// concern, it's trivial to write more sophisticated implementations, since the logic is entirely
+// encapsulated.
+//
+// The class is current a subclass of HashObject internally, but an Object direct subclass at
+// user level.
+type ConcurrentHashObject struct {
+	*baseObj
+	InternalHash *HashObject
+
+	sync.RWMutex
```
st0012: You should not embed the original HashObject with lock. This performs poorly than using `sync.Map` as data storage directly. See this:


[Reference](https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c)

vm/vm.go
```diff
@@ -212,6 +212,7 @@ func (vm *VM) initConstants() {
 		vm.initNullClass(),
 		vm.initArrayClass(),
 		vm.initHashClass(),
+		vm.initConcurrentHashClass(),
```
st0012: I think it should be a standard library that need to be required when needed. If so, you should put the initialize method here:

```go
var standardLibraries = map[string]func(*VM){
	"net/http":          initHTTPClass,
	"net/simple_server": initSimpleServerClass,
	"uri":               initURIClass,
	"db":                initDBClass,
	"plugin":            initPluginClass,
	"json":              initJSONClass,
        // Here
        "concurrent_hash": vm.initConcurrentHashClass,
}
```

Note that the builtin classes and standard library class's initialize method is slightly different.

vm/concurrent_hash.go
```diff
@@ -0,0 +1,377 @@
...
+						return t.vm.initConcurrentHashObject(hashArg.Pairs)
+					}
+				}
+			},
+		},
+	}
+}
+
+// Instance methods -----------------------------------------------------
+func builtinConcurrentHashInstanceMethods() []*BuiltinMethodObject {
```
st0012: @saveriomiroddi  Are these methods' implementation same as the original Hash? If so, how about make

```go
h := receiver.(*ConcurrentHashObject)
// or
h := receiver.(*HashObject)
```
become something like:

```go
type hash type {
    load() Object, bool
    store(Object)
}

h := receiver.(hash)
v, ok := h.load()
```

Then we can just implement the interface on both `HashObject` and `ConcurrentHashObject`. After that we can use same method implementations for both classes. (This could be a large refactoring though)

saveriomiroddi: `sync.Map` has some API and semantic differences; the generic associative array functionality can be abstracted, but not completely (at least, not completely and exactly).

Have a look at the `sync.Map` documentation and let me know, since some design decisions need to be taken, in particular, due the non-determinism of the iteration.

st0012: I know `sync.Map` has its own API, so what I suggest is to wrap another layer around both normal map and `sync.Map`'s APIs. For example:

```go
func (*ch ConcurrentHashObject) load(key *StringObject) Object {
    value, ok := h.internalMap.Load(key.value)
    if !ok {
        return NULL
    }
   return value.(Object)
}

func (*h HashObject) load(key *StringObject) Object {
    value, ok :=h.Pairs[key.value]
    if !ok {
        return NULL
    }
   return value
}
```

Then if we implement all the Hash methods using those interface APIs, we can share those methods between these two types.

saveriomiroddi: The problems I refer to are:

1. iteration: supported with different API
2. length: it's not supported at all by sync.Map

Problem 1. can be solved easily, although a bit clunkily, by abstracting a `range()` function, and having Hash methods pass a function to it, instead of using the built-in `range` keyword

Problem 2. is significant, as it has both functional and semantic implications. the functional problem can be solved (if necessary) through count via iteration (which will be in linear rather than constant time) . the semantic aspect is something that needs to be explicitly accepted or rejected; if accepted, it needs to be documented very clearly that iteration (and count) will be non-deterministic.

st0012: I think I'll drop the `length` method from concurrent map.

I just took a look at [this issue](https://github.com/golang/go/issues/20680) and there's no clear usage for the `Len` method of `sync.Map`. So I assume there won't be much usage for concurrent hash's `length` method either. And user can use iteration to count the number themselves if needed.

saveriomiroddi: :+1:

saveriomiroddi:
@st0012 ok, I've applied all the changes required, so it's ready for review.

vm/concurrent_hash.go
saveriomiroddi: I know that article :wink:; it's import to have clear that it's not meant to be used as a reference for the general speed of `sync.Map` (it's actually explicitly written both in the article, and in the sync.Map documentation).

`sync.Map` is designed with a specific case in mind (essentially, high volume reads with sparse writes), and the article tests that case only, in ideal conditions. Under other circumstances (for example, certain caching patterns), `sync.Map` will have [comparable or worse performance](https://golang.org/pkg/sync/#Map).

With this in mind, since the Goby implementation targets no specific use case, until there is no real-world usage, both are equally valid.  

Probably a good idea would be to provide Goby mutex APIs, so that if users want to implement a lock-based hash, they can.

st0012: @saveriomiroddi Sorry for not completely understanding the reference and provided wrong suggestion 😢. I think I should take a time to have a deeper understanding on concurrent programming. Do you think [this](https://stackoverflow.com/questions/19148809/how-to-use-rwmutex-in-golang#answer-19168242) is a clear and correct explanation of Go's `sync.Mutex` and `sync.RWMutex`?

vm/db_test.go
```diff
@@ -51,7 +47,7 @@ func TestPGConnectionPing(t *testing.T) {
 
 	for i, tt := range tests {
 		v := initTestVM()
-		evaluated := v.testEval(t, tt.input, getFilename())
+		evaluated := v.testEvalWithRequire(t, tt.input, getFilename(), "db")
```
st0012: I'm not quite sure about this change actually. Here's my opinions of it:

1. This deserves another PR, because it's not really related to the subject.
2. I think keep those `require`s in the script is more easy for others to understand how it works. They just need to take a look at the script itself instead of both script and the `v.testEvalWithRequire` function's argument.

BTW the `testEvalWithRequire` only takes one library's name, what if I want to require multiple libraries in my test script? Shouldn't the function be able to require multiple libraries?

Anyway, I personally don't think this is a necessary change.

@saveriomiroddi Can you explain about the reason you want to make this change? Is it because those require statement against the DRY principle?

@goby-lang/maintainers How do you think about this change?

vm/concurrent_hash_test.go
```diff
@@ -0,0 +1,586 @@
...
+		return false, err
+	}
+	if err := json.Unmarshal(b, &j2); err != nil {
+		return false, err
+	}
+	return reflect.DeepEqual(j2, j), nil
+}
+
+// We can't compare string directly because the key/value's order might change and we can't control it.
+func ConcurrentHashcompareJSONResult(t *testing.T, evaluated Object, exp interface{}) {
```
st0012: Same as above

vm/concurrent_hash_test.go
```diff
@@ -0,0 +1,586 @@
...
+	for i, tt := range testsFail {
+		v := initTestVM()
+		evaluated := v.testEvalWithRequire(t, tt.input, getFilename(), "concurrent_hash")
+		checkError(t, i, evaluated, tt.expected, getFilename(), tt.errorLine)
+		v.checkCFP(t, i, 1)
+		v.checkSP(t, i, 1)
+	}
+}
+
+func ConcurrentHashJSONBytesEqual(a, b []byte) (bool, error) {
```
st0012: I think this can be removed, right?

vm/classes/classes.go
```diff
@@ -1,22 +1,23 @@
...
-	MatchDataClass = "MatchData"
-	GoMapClass     = "GoMap"
+	ObjectClass         = "Object"
+	ClassClass          = "Class"
+	IntegerClass        = "Integer"
+	FloatClass          = "Float"
+	StringClass         = "String"
+	ArrayClass          = "Array"
+	HashClass           = "Hash"
+	ConcurrentHashClass = "ConcurrentHash"
```
st0012: `ConcurrentHash` is kind of a library, so it shouldn't have a constant here. Here's only for builtin classes (sorry I should clarify this in comment).

st0012: And I prefer `Concurrent::Hash` over `ConcurrentHash`, are you ok with this?

vm/concurrent_hash.go
saveriomiroddi: > Sorry for not completely understanding the reference and provided wrong suggestion

No problem :smile: 

Actually, I think that by providing synchronization APIs (#477), the new implementation will give users (programmers) flexibility for all the use cases:

- ConcurrentHash, based on sync.Map, will provide optimal performance for the mentioned use case
- regular Hash + synchronization APIs, for all the use cases where ConcurrentHash doesn't perform

With the previous implementation (ConcurrentHash based on RWLock), there were choice for users (programmers).

I think we can follow up with the discussion on #477.

Very good! :+1: 

> Do you think this is a clear and correct explanation of Go's sync.Mutex and sync.RWMutex

I think so, yes!

vm/classes/classes.go
saveriomiroddi: Makes sense, I will change it :+1:

vm/concurrent_hash_test.go
saveriomiroddi: :+1:

vm/concurrent_hash_test.go
saveriomiroddi: :+1:

vm/db_test.go
saveriomiroddi: > Can you explain about the reason you want to make this change? Is it because those require statement against the DRY principle?

The reason was DRY, correct - for each test file, the `require` is repeated over and over.
It is obviously a requirement for the tests to run, but in a way, from the point of view of each unit test, the `require` statement itself is irrelevant.

> BTW the testEvalWithRequire only takes one library's name, what if I want to require multiple libraries in my test script?

It's a valid point. Since all the UTs always use at most one require, I've implemented a single parameter, but when/if somebody will need multiple requires, variadic args could be used.

> This deserves another PR, because it's not really related to the subject.

Makes sense. I'll remove this commit, and open an issue - then the issue can be discussed and implemented or closed.

saveriomiroddi:
@st0012 review changes applied.

st0012:
@saveriomiroddi Can you fix the second issue? The first one is on me so I'll fix it after merge this.
https://www.codacy.com/app/stan001212/goby/pullRequest?prid=1010027

saveriomiroddi:
>  Can you fix the second issue? The first one is on me so I'll fix it after merge this.

@st0012  done :smile:

st0012:
@saveriomiroddi Nice work, thanks!!

PR #475, opened 2017-10-23
Error stacktraces

### LINKED ISSUE #352: Better Error report
Currently we can see where causes a error, but only for Goby code. If a error happens in a builtin (written in Go) method's block argument then we'd never know what's going on.

st0012:
Sample code:

```ruby
class Car
  def initialize
    yield(self)
  end

  def color=(c)
    @color = c
  end

  def doors=(ds)
    @doors = ds
  end
end

car = Car.new do |c|
  c.color = "Red"
  c.doors = 4
  foo # Error
end
```

And its stack traces will be:
```
UndefinedMethodError: Undefined Method 'foo' for <Instance of: Car>
from /Users/stanlow/projects/go/src/github.com/goby-lang/goby/samples/sample-6.gb:18
from /Users/stanlow/projects/go/src/github.com/goby-lang/goby/samples/sample-6.gb:15
```

Closes #352

### DIFF
--- a/vm/error_test.go
+++ b/vm/error_test.go
@@ -179,6 +179,7 @@ func TestStackTraces(t *testing.T) {
 		input          string
 		expectedMsg    string
 		expectedTraces []string
+		expectedCFP    int
 	}{
 		{`def foo(a, b, c)
 		  a + b + c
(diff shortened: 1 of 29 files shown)


saveriomiroddi:
@st0012 Hello! Should I review it now or wait for it to be finished? I see some [test] code is commented.

vm/error_test.go
```diff
@@ -148,15 +155,159 @@ func TestArgumentError(t *testing.T) {
...
 		evaluated := v.testEval(t, tt.input, getFilename())
-		checkError(t, i, evaluated, tt.expected, getFilename(), tt.errorLine)
+		checkErrorMsg(t, i, evaluated, tt.expected)
 		v.checkCFP(t, i, tt.expectedCFP)
-		v.checkSP(t, i, 1)
+		v.checkSP(t, i, tt.expectedSP)
+	}
+}
+
+func TestStackTraces(t *testing.T) {
```
st0012: @goby-lang/maintainers We should have error's stack traces in the future! There's still some places need to be improved, but at least the basic functionality is done.

st0012:
@hachi8833 @saveriomiroddi @Maxwell-Alexius I hope you can all take a look if you have time. Mainly on `vm/error.go`, `vm/error_test.go` and `vm/thread.go`.

PR #483, opened 2017-10-27
Decouple instructions

st0012:
In the last refactoring I pass instructions to every instruction's operations and error initialization function. But it turns out we only need its `sourceLine (int)` attribute. So I decide to just pass the attribute (`sourceLine` in this case) we need instead of the instruction pointer, which increased coupling.

If in the future we need more info when executing instruction's operation, we should have a specialized data structure to store it.


vm/thread.go
```diff
@@ -162,7 +162,7 @@ func (t *thread) retrieveBlock(fileName, blockFlag string) (blockFrame *normalCa
 	return
 }
 
-func (t *thread) sendMethod(methodName string, argCount int, blockFrame *normalCallFrame, instruction *instruction) {
+func (t *thread) sendMethod(methodName string, argCount int, blockFrame *normalCallFrame, sourceLine int) {
```
st0012: @saveriomiroddi You can use this function to call Goby method inside Go. You should pass the method's name and the number of arguments you provided. Leave `blockFrame` to `nil`, and `sourceLine` to be `-1`.

And the most important thing is you need to push every arguments your want to use on the stack in order. This can't be simplify because builtin method and normal methods have different ways to retrieve arguments object from stack.

PR #490, opened 2017-10-29
[WIP] ConcurrentArray implementation via inheritance

saveriomiroddi:
@st0012 this is the alternative implementation I've mentioned in the other ConcurrentArray PR; I've implemented only one method, for showcasing the idea.

The second commit shows how the arguments checking should be performed outside of the thread-protected method, since checking doesn't need any locking.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -56,10 +56,21 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
 			Name: "[]",
 			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
 				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
+					if len(args) != 1 {
+						return t.vm.initErrorObject(errors.ArgumentError, sourceLine, "Expect 1 arguments. got=%d", len(args))
+					}
+
+					i := args[0]
+					index, ok := i.(*IntegerObject)
+
+					if !ok {
+						return t.vm.initErrorObject(errors.TypeError, sourceLine, errors.WrongArgumentTypeFormat, classes.IntegerClass, args[0].Class().Name)
+					}
+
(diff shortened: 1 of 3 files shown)


st0012:
@saveriomiroddi This is pretty close to what I thought, and it's more like `concurrent-ruby`'s `Concurrent::Array` implementation I think. Which approach do you prefer?

From my opinion, the main consideration will be if we would need to customize concurrent array object's methods in the future. For example, use `atomic` package to do more low-level lock operations to improve performance (although I don't think we'll have to do this).

saveriomiroddi:
I've checked the concurrent-ruby project.  
Technically speaking, their solution is very similar to #487 - both are wrappers introducing a mutex (see https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/array.rb and https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/thread_safe/util/array_hash_rbx.rb). The difference with this PR is due to the fact that here the superclass (top-level Array) is modified, while in ruby's concurrent, and #487, we don't.

From a theoretical perspective, this solution will introduce a small overhead in ::Array methods, due to calling the locking methods (which will be both empty), and it will also of course introduce the convention to use in such class (that is, defining the code a `func`).
The first may be negligible, the second is a design choice.

I'm fine with this approach, also because it should work out of the box.

st0012:
@saveriomiroddi Um...After comparison I think your design (#487) is more straightforward and easy to understand. So if you don't mind I'll close this and we can go back to #487. I will help you solve the CFP issue.

saveriomiroddi:
> Um...After comparison I think your design (#487) is more straightforward and easy to understand. So if you don't mind I'll close this and we can go back to #487. I will help you solve the CFP issue.

:+1: Closing.

PR #501, opened 2017-11-04
Add `Object#object_id`

hachi8833:
```ruby
» true.object_id
#» 842351475936
» true.object_id
#» 842351475936
» String.object_id
#» 842351474240
» "abc".object_id
#» 842352149632
» 1.object_id
#» 842352567584
» a = "3.14".to_f
#» 3.14
» a.object_id
#» 842352568576
```

### DIFF
--- a/vm/class_test.go
+++ b/vm/class_test.go
@@ -996,11 +996,27 @@ func TestObjectId(t *testing.T) {
 		input    string
 		expected interface{}
 	}{
+		// immutable objects
+		{`Object.new.object_id.is_a?(Integer)`, true},
 		{`a = nil.object_id; b = nil.object_id; a == b`, true},
 		{`a = true.object_id; b = true.object_id; a == b`, true},
 		{`a = false.object_id; b = false.object_id; a == b`, true},
+		{`CONSTANT=1; CONSTANT.object_id == CONSTANT.object_id`, true},
+		{`Class.object_id == Class.object_id`, true},
+		{`Object.object_id == Object.object_id`, true},
+		{`Integer.object_id == Integer.object_id`, true},
+		// mutable objects
 		{`a = 1.object_id; b = 1.object_id; a == b`, false},
 		{`a = "a".object_id; b = "a".object_id; a == b`, false},
+		{`a = 1.object_id; b = a; a.object_id == b.object_id`, true},
+		{`a = "a".object_id; b = a; a.object_id == b.object_id`, true},
+		{`a = 'a'.object_id; b = a; a.object_id == b.object_id`, true},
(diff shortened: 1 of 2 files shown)


vm/class_test.go
```diff
@@ -990,3 +990,24 @@ func TestClassSingletonClassClassMethod(t *testing.T) {
...
+func TestObjectId(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		{`a = nil.object_id; b = nil.object_id; a == b`, true},
+		{`a = true.object_id; b = true.object_id; a == b`, true},
+		{`a = false.object_id; b = false.object_id; a == b`, true},
+		{`a = 1.object_id; b = 1.object_id; a == b`, false},
+		{`a = "a".object_id; b = "a".object_id; a == b`, false},
```
hanachin: I looked ruby/spec's `#object_id` spec, we should add test for the `@return [Integer]`
https://github.com/ruby/spec/blob/e84c928a9ce171693a1221da88cf98e3837a77fa/shared/kernel/object_id.rb
```rb
{`Object.new.object_id.is_a?(Integer)`, true}
```

hanachin: There are serveral built-in classes which have the literal syntax, Should we have test cases for them?
e.g. `24..7`, `{}`, `[]`, `:sym`
https://github.com/goby-lang/goby/blob/cc8f18446d36d19945f97a1de9000f79c9bc6558/compiler/parser/data_type_parsing.go

hachi8833: Added

hachi8833: Added. Thank you!

vm/class_test.go
```diff
@@ -990,3 +990,40 @@ func TestClassSingletonClassClassMethod(t *testing.T) {
...
+		{`Object.new.object_id.is_a?(Integer)`, true},
+		{`a = nil.object_id; b = nil.object_id; a == b`, true},
+		{`a = true.object_id; b = true.object_id; a == b`, true},
+		{`a = false.object_id; b = false.object_id; a == b`, true},
+		{`CONSTANT=1; CONSTANT.object_id == CONSTANT.object_id`, true},
+		{`Class.object_id == Class.object_id`, true},
+		{`Object.object_id == Object.object_id`, true},
+		{`Integer.object_id == Integer.object_id`, true},
+		// mutable objects
+		{`a = 1.object_id; b = 1.object_id; a == b`, false},
```
st0012: I don't think this means objects are mutable, we just create a new one every time.

hachi8833: I'll fix this.

PR #502, opened 2017-11-05
[Experimental] Decimal class

hachi8833:
I have implemented a `Decimal` class by using Go's math/big package, just to see if the class works or not. Currently we have no float/decimal literal and you need `String#to_d` to create Decimal object.

Internally, the Decimal is a rational number and consists of a numerator and a denominator.

```ruby
"3.14".to_d            # => 3.14
"-0.7238943".to_d      # => -0.7238943
"355/113".to_d         # => 3.1415929203539823008849557522123893805309734513274336283185840

a = "16.1".to_d
b = "1.1".to_d
e = "17.2".to_d
a + b # => 0.1
a + b == e # => true

('16.1'.to_d  + "1.1".to_d).to_s #=> 17.2
('16.1'.to_f  + "1.1".to_f).to_s #=> 17.200000000000003
```

- [x] `#+`
- [x] `#-`
- [x] `#*`
- [x] `#/`
- [x] `#**`
- [x] `#>`
- [x] `#<`
- [x] `#>=`
- [x] `#<=`
- [x] `#==`
- [x] `#!=`
- [x] `#<=>`
- [x] `#fraction`
- [x] `#inverse`
- [x] `#denominator`
- [x] `#numerator`
- [x] `#reduction`
- [x] `#to_a`
- [x] `#to_f`
- [x] `#to_i`
- [x] `#to_s`
- [x] `String#to_d`
- [x] `Float#to_d`

If the idea or implementation does not fit to @st0012's  philosophy, I'm ok to discard this PR.


st0012:
@hachi8833 We'll have float literal after @Maxwell-Alexius finish #470, but I guess the decimal still won't have its own literal?
And I think you should also add `#to_d` to `Float`

hachi8833:
Actually, it is pretty hard for me to add a literal so far. 
I'd check adding `Float#to_d`.

st0012:
@hachi8833 Um...I'm not saying we need a decimal literal. Instead, I think we might not need this?

hachi8833:
I got it 💦 Just adding `Float#to_d` might be sufficient so far.

st0012:
LGTM. @saveriomiroddi @Maxwell-Alexius Can you also help me review this?
BTW @Maxwell-Alexius what blocks you from finishing #470?

vm/decimal_test.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+}
+
+func TestDecimalEquality(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		{`'123.5'.to_d  ==  '123.5'.to_d`, true},
+		//{`'123'.to_d    ==  123`, true},
```
st0012: Why comment out this case?

vm/decimal_test.go
```diff
@@ -0,0 +1,388 @@
...
+		v.checkSP(t, i, 1)
+	}
+}
+
+func TestDecimalConversions(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		//{`'100.3'.to_d.to_i`, 100},
```
st0012: Here as well

vm/string.go
```diff
@@ -1452,6 +1453,40 @@ func builtinStringInstanceMethods() []*BuiltinMethodObject {
...
+			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
+
+					if len(args) != 0 {
+						return t.vm.initErrorObject(errors.ArgumentError, sourceLine, "Expect 0 argument. got=%v", strconv.Itoa(len(args)))
+					}
+
+					str := receiver.(*StringObject).value
+
+					re := regexp2.MustCompile("[^0-9./\\-]", 0)
```
st0012: I think we can remove this section, just check it when calling `SetString`.

vm/decimal_test.go
hachi8833: I'll open it

vm/decimal_test.go
hachi8833: I'll open it

st0012:
@hachi8833 I took a closer look and can't understand the usage of these APIs:

- `#to_ai`
- `#to_h`
- `#to_hi`

Can you explain why you add these APIs?

hachi8833:
Thank you for pointing them out. I realized the three are redundant.

PR #506, opened 2017-11-11
Fix description for `-e` option

kachick:


hachi8833:
Thank you for the PR, too!
Now I found that `-e` option causes a panic; I will open another issue for that.

kachick:
Thank you!

>Now I found that -e option causes a panic; I will open another issue for that.

😮 🙏

Issue #509, opened 2017-11-11
Bug: `-e` option causes a panic


hachi8833:
`-e` option for starting up Goby causes a panic.

- Your operating system: Mac OS X High Sierra
- How did you install Goby: git clone
- Goby version: 0.1.3
- The value of `$GOBY_ROOT`: /Users/hachi8833/deve/golang/gopath/sys/src/github.com/goby-lang/goby

```ruby
# report.gb
puts "hello"
```

```bash
goby -e report.gb
Will generate issue report on error...

hello
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x43ea690]

goroutine 1 [running]:
github.com/goby-lang/goby/vm.PrintError(0xc4200c42c0)
	/Users/hachi8833/deve/golang/gopath/sys/src/github.com/goby-lang/goby/vm/issue_vm.go:20 +0x50
main.main()
	/Users/hachi8833/deve/golang/gopath/sys/src/github.com/goby-lang/goby/goby.go:93 +0x4d6
```

PR #511, opened 2017-11-11
#509 Fix panic on `-e` option

hachi8833:
To fix https://github.com/goby-lang/goby/issues/509


st0012:
@goby-lang/maintainers  Not directly related to this PR, but I think we should start writing tests for `goby` command and flags. Is there anyone want to do this?
@hachi8833 Can you help me open an issue and reference to here?

PR #512, opened 2017-11-11
Implement Module(Class)#>

kachick:
If the direction is not wrong, I can add `#<` , `#>=`, `#<=` in other PR .

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -94,7 +94,7 @@ func builtinClassCommonClassMethods() []*BuiltinMethodObject {
 			// ```
 			//
 			// @param module [Class]
-			// @return [Boolean]
+			// @return [Boolean, Null]
 			Name: ">",
 			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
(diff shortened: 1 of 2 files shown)


hachi8833:
LGTM.
@st0012 How about this?

st0012:
@kachick @hachi8833 I think this could be added, but we'll need more complex test cases. For example: 

```ruby
module Foo; end
class Bar; end

Object > Foo #=> nil
Object < Foo #=> nil
Bar > Foo #=> false
Bar < Foo #=> true
```

st0012:
Can you find related ruby specs from [Ruby Spec](https://github.com/ruby/spec)?

kachick:
Of sorry, perfectly forgotten to compare with non related modules 🙇 
Applied it in https://github.com/goby-lang/goby/pull/512/commits/5417218a5db6d48c4fa9968ac53ebebfa61fbd7b

kachick:
Thank you! ☺️

Issue #508, opened 2017-11-11
Nil groups make panic in MatchData


kachick:
- Your operating system: `sw_vers`
  - ProductName:    Mac OS X
  - ProductVersion: 10.13.1
  - BuildVersion:   17B48
- How did you install Goby: `go get github.com/goby-lang/goby`
- Goby version: 0.1.3
- The value of `$GOBY_ROOT`: /Users/kachick/gocode/src/github.com/goby-lang/goby

```ruby
'b'.match(Regexp.new('(a)?'))
```

```
panic: runtime error: index out of range
```

- Error messages

```
# goby -i
Goby 0.1.3 👺 🤠 😎
» 'b'.match(Regexp.new('(a)?'))
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/goby-lang/goby/vendor/github.com/dlclark/regexp2.(*Capture).String(...)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/match_data.go:123
github.com/goby-lang/goby/vm.(*VM).initMatchDataObject(0xc4200b42c0, 0xc4201da000, 0xc42019c094, 0x4, 0xc42019c048, 0x1, 0x0)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/match_data.go:123 +0x506
github.com/goby-lang/goby/vm.builtinStringInstanceMethods.func28.1(0xc4201360f0, 0xc42019e390, 0x1, 0x1, 0x0, 0xc42019e390, 0x0)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/string.go:1063 +0x107
github.com/goby-lang/goby/vm.(*thread).evalCallFrame(0xc4201360f0, 0x480d1c0, 0xc4201c2360)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/thread.go:63 +0x214
github.com/goby-lang/goby/vm.(*thread).startFromTopFrame(0xc4201360f0)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/thread.go:42 +0x61
github.com/goby-lang/goby/vm.(*thread).evalBuiltinMethod(0xc4201360f0, 0x480c8a0, 0xc4201c2200, 0xc4200ef980, 0x0, 0x1, 0xc4201be3c0, 0x0, 0x1, 0x4565223, ...)
        /Users/kachick/gocode/src/github.com/goby-lang/goby/vm/thread.go:240 +0x3f8
[...]

PR #513, opened 2017-11-11
Refactor: Add type alias for regexp2.Regexp

hachi8833:
- Add a type alias `Regexp` for `*regexp2.Regexp`
- Change the attribute `Regexp` to the lowercase `regexp`


st0012:
@hachi8833 Um....I don't think this refactoring is necessary. Can you explain you intention? Maybe I missed something.

hachi8833:
Sorry for the short explanation. 
The refactoring itself is not so important, but it is actually a small preparation of fixing `MatchData`.

I checked the bug https://github.com/goby-lang/goby/issues/508 and found that `MatchData` needs fix. But I also found that current struct in `MatchData` is custom-made; I think `*regexp.Match` should be used instead and now working on it.

I close this for now and send another PR for fixing `MatchData` including this refactoring. Just a moment.

st0012:
@hachi8833 I see

PR #517, opened 2017-11-14
Add checks the number of args in `require` and `require_relative`

hachi8833:
- Add a check for the number of arguments in `require` and `require_relative`
- Update the API document

FYI: I happened to find that `require :db` works. Looks like this comes from the fact that symbols are automatically converted into strings. Of course `require :net/http` does not work.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -855,15 +855,8 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 			// Loads the given Goby library name without extension (mainly for modules), returning `true`
 			// if successful and `false` if the feature is already loaded.
 			//
-			// Currently, only the following embedded Goby libraries are targeted:
-			//
-			// - "file"
-			// - "net/http"
-			// - "net/simple_server"
-			// - "uri"
-			//
 			// ```ruby
-			// require("file")
+			// require("db")
 			// File.extname("foo.rb")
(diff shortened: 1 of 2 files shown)


st0012:
@hachi8833  LGTM
And the symbol part you mentioned it's worth recorded, but we can ignore that for now.

saveriomiroddi:
Note: if it's not possible to internally distinguish between symbols and strings, then ignore the comment about :smile:

st0012:
@saveriomiroddi Currently it's impossible to know if the string object comes from symbol or normal string. So this can't be forbidden now.
BTW, I think this is a good time to talk about string interning. I always want to implement string interning with only `String` type, while Ruby does this with symbol.
The concern is that string interning is like a hash table for string/string object's pointer, and every time we create/return a string object need to check that table first. This will be a big issue when running multi-threaded programs, because every different thread might need to access/write that table. And we can't come up with an elegant solution, we need to lock that table, which will cause poor performance.

saveriomiroddi:
It seems JRuby, which doesn't have a GIL, and whose core is thread-safe, performs strings interning:

- https://stackoverflow.com/a/44637983
- https://github.com/jruby/jruby/issues/3491

so it may be worth looking at how JRuby tackles this problem (I'm not 100% sure it does, but I think so).

PR #519, opened 2017-11-16
Fix travis build running twice the test suite

### LINKED ISSUE #518: Test suite is run twice
The travis build script has a bug, that causes the build to run twice. This was hidden because for the first run, the verbose output was disabled, so there was only one line report line for it.

saveriomiroddi:
Fixes #518

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -11,7 +11,7 @@ for d in $(go list ./...); do
         # This can generate full coverage report of vm package.
         # Test that need to run without race detection include NoRaceDetection in the name,
         # otherwise, they will run twice (in the run below).
-        TEST_PLUGIN=true go test -coverprofile=profile.out -covermode=atomic $d -v -run NoRaceDetection
+        NO_RACE_DETECTION=true go test -coverprofile=profile.out -covermode=atomic $d -v -run NoRaceDetection
         if [ -f profile.out ]; then
           cat profile.out >> coverage.txt
           rm profile.out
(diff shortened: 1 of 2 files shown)


st0012:
@saveriomiroddi It seems that the new script doesn't import test coverage data properly, so the [coverage](https://codecov.io/gh/goby-lang/goby) is now down to 32%.

st0012:
I should've noticed this before merge. I thought the 32% is about target coverage 😫

saveriomiroddi:
> It seems that the new script doesn't import test coverage data properly, so the coverage is now down to 32%.

This is very likely caused by the original script having a bug also in the coverage calculation; in this case, it should be because the strategy used:

    if [ -f profile.out ]; then
      cat profile.out >> coverage.txt
      rm profile.out
    fi

doesn't work as intended, for some reason.

st0012:
Um.....this should be fixed ASAP. I'm investigating now.

saveriomiroddi:
It's hard to work on this area (full testing) without the database setup. I'll work on setting up the database next, and I'll put the instructions on the wiki.

st0012:
If you already have PG installed and have a proper PG role/user, run these commands and it should work:

```
psql
create database goby_test;
```

saveriomiroddi:
During the last attempt, things didn't turn out to be trivial overall, I'll need to review it.

st0012:
I think I found the reason, the command that runs all vm tests doesn’t generate profile result.
We used to use the coverage of the first run and never have the coverage of the second run. And the old first run script doesn’t have the —run flag to filter test cases, so it runs all tests and gives full coverage. But now it only runs those not race detecting tests.
So the solution should be adding the same if statement block that inserting profiling results, just below the second vm test command and above continue (also remember to add coverage flags to second test command)

PR #520, opened 2017-11-16
Add MatchData#to_h

hachi8833:
dlclark/regexp2 has an ability to handle named-capture, so I added `to_h` to obtain a hash with named capture.


st0012:
@hachi8833 Wow the `regexp2` package is really powerful.

PR #522, opened 2017-11-17
Fix VM test case's coverage calculation

st0012:


st0012:
@saveriomiroddi The issue seems to be fixed, can you help review it?

PR #536, opened 2017-12-12
`Range#map`

hachi8833:
Added `Range#map` as `Array#map`.

```ruby
(1..10).map do |i|
 i * i
end
# => [1, 4, 9, 16, 25, 36, 49, 64, 81, 100]
```


st0012:
@hachi8833 Can you update the branch and fix the spec?

PR #542, opened 2017-12-23
Rename verify functions

st0012:
We used to use `testXXX` or `checkXXX` as our verifying function's name, I want to make them have `verify` prefix instead (or `assert`?)


st0012:
@hachi8833 Can you use `approve` if you think it's ok next time? thx

hachi8833:
I got it!

PR #543, opened 2017-12-24
Add a compile option to reduce binary size

hachi8833:
Just a small change to reduce Goby's binary size.


st0012:
@hachi8833 I remember we can use compiled binary to run tests in Go, can you help checking that? If so, I want to use the reduced sized binary to run tests

hachi8833:
The option to reduce binary size works on `make test`.

st0012:
@hachi8833 Can you apply these flags into `travis.sh`'s test commands?

hachi8833:
I did. I hope this works.

hachi8833:
I think CI runs faster now. Merging...

Issue #548, opened 2017-12-26
`if` syntax with `then` causes interpreter panic


saveriomiroddi:
Using `if` with `then` causes the interpreter to panic:

    Goby 0.1.6 😶 😈 👻
    » if true then
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x74ece5]

    goroutine 1 [running]:
    github.com/goby-lang/goby/compiler/ast.(*BlockStatement).KeepLastValue(0xc42015af80)
      /path/to/go/src/github.com/goby-lang/goby/compiler/ast/statements.go:215 +0x65
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseConditionalExpression(0xc42011a1b0, 0x0)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/flow_control_parsing.go:120 +0x15e
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseConditionalExpressions(0xc42011a1b0, 0xc4201722d0, 0x0, 0x0)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/flow_control_parsing.go:106 +0x40
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseIfExpression(0xc42011a1b0, 0xbc7688, 0xc42011d770)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/flow_control_parsing.go:92 +0xac
    github.com/goby-lang/goby/compiler/parser.(*Parser).(github.com/goby-lang/goby/compiler/parser.parseIfExpression)-fm(0xb3c220, 0xc420172060)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/parser.go:143 +0x2a
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseExpression(0xc42011a1b0, 0x2, 0x8, 0x18)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/expression_parsing.go:133 +0xbd
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseExpressionStatement(0xc42011a1b0, 0x4)
      /path/to/go/src/github.com/goby-lang/goby/compiler/parser/statement_parsing.go:271 +0x146
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseStatement(0xc42011a1b0, 0x10c9a40, 0x1)
[...]

PR #551, opened 2017-12-27
Drop support of then keyword

amohamedali:


st0012:
@amohamedali Just a reminder: Remember to add `This closes issue #ISSUE_NUMBER` in the description for automatically close related issues.

amohamedali:
Oops my bad I will added from now on. 
This closes issue #548 

Otherwise it seems strange to raise an unexpected token 'then' as 'then' now on doesn't become a "token" no ?

st0012:
@amohamedali You should let it return the same kind of error message as following code

```ruby
case int
when 0 foo
  foo
end
```

I believe that'll be something like `unexpected token Ident 'foo'`, and in this case it should be `unexpected token Ident 'then'`.

And you should edit this PR's description and put the closing issue line in it instead of in the comment.

PR #552, opened 2017-12-28
While statement require a do keyword after conditions

amohamedali:


hachi8833:
Thank you for the PR!
I think we need some more evaluation test on vm/evaluation_test.go for success/fail in `while` and `do` statements.

Any suggestions? @st0012

st0012:
@hachi8833 We already have vm's success tests for `while` statement. And if we are talking about the syntax failure, I don't think we should test that in vm

st0012:
@amohamedali great job! thx 😄

Issue #545, opened 2017-12-24
Question: Behavior of `Hash#map_value`


hachi8833:
I'm wondering the behavior of `Hash#map_value`:

* While the API says "This method does not change the keys and the receiver hash values", this method actually changes the receiver.
* The actual return value is `Hash` while API doc says `Boolean`
* Using `h` and `result` looks redundant: using only `h` is sufficient.

```go
			// Returns a new hash with the results of running the block once for every value.
			// This method does not change the keys and the receiver hash values.
			//
			// ```Ruby
			// h = { a: 1, b: 2, c: 3 }
			// result = h.transform_values do |v|
			//   v * 3
			// end
			// h      # => { a: 3, b: 6, c: 9 }
			// result # => { a: 3, b: 6, c: 9 }
			// ```
			//
			// @return [Boolean]
			Name: "map_values",
			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
					if len(args) != 0 {
						return t.vm.initErrorObject(errors.ArgumentError, sourceLine, "Expect 0 argument. got: %d", len(args))
					}

					if blockFrame == nil {
						return t.vm.initErrorObject(errors.InternalError, sourceLine, errors.CantYieldWithoutBlockFormat)
					}

					h := receiver.(*HashObject)

					if len(h.Pairs) == 0 {
						t.callFrameStack.pop()
					}

					for k, v := range h.Pairs {
						result := t.builtinMethodYield(blockFrame, v)
						h.Pairs[k] = result.Target
					}
					return h
				}
			},
```

Could you please check this? > @Maxwell-Alexius

Maxwell-Alexius:
Thanks, @hachi8833, I think I can handle this

hachi8833:
Thanx!

hachi8833:
Fixed on https://github.com/goby-lang/goby/pull/553

PR #553, opened 2017-12-29
#545 Fix Hash#map_values Behaviour

Maxwell-Alexius:
@hachi8833 Now, I've changed the method to the correct behaviour which, as you've mentioned, the receiver (the hash) itself shouldn't change the value. Instead, it returns the new one.

However, I create another variable `result` is for not changing the original receiver `h` because it is passed by reference (I think?) so if changing the value in `h` itself, it will also alter the value of the receiver.

This PR should solve issue #545

### DIFF
--- a/vm/hash.go
+++ b/vm/hash.go
@@ -825,10 +825,10 @@ func builtinHashInstanceMethods() []*BuiltinMethodObject {
 			//
 			// ```Ruby
 			// h = { a: 1, b: 2, c: 3 }
-			// result = h.transform_values do |v|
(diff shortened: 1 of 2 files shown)


Maxwell-Alexius:
BTW, the correct behaviour should be something like this:

```Ruby
hash = { a: 1, b: 2, c: 3 }
result = hash.map_values do |v|
  v * 3
end

hash # => { a: 1, b: 2, c: 3 }
result # => { a: 3, b: 6, c: 9 }
```

Issue #555, opened 2017-12-30
Positive unary operator like `+1` unhandled


hachi8833:
When fixing #526, I also found that the positive unary operator like `+1` are treated as invalid:

```ruby
» -1
#» -1
» +1
unexpected + Line: 0
» +1
```

I'd like to fix this as well when fixing #526 and then close this issue.

st0012:
ok

st0012:
@hachi8833 Has this been finished? I remember you almost finished it few weeks ago

hachi8833:
First I thought fixing #568 together, but I realized this should be fixed in another PR.
This will be finished this week, too. Just a moment.

Issue #554, opened 2017-12-30
Remove unused Incr ++ and Decr ++


hachi8833:
While fixing #526, I found that `token.Incr` and `token.Decr` are used in lexer.go while increment/decrement operators are unused.

```go
// compiler/token/token.go
	Incr     = "++"
	Decr     = "--"
```

I'd like to remove them when fixing  #526. If @st0012 has any plan to use them, please let me know.

st0012:
@hachi8833 They will not be used again, plz drop them, thx

Issue #526, opened 2017-11-18
Fix Negative Float Type Data Parsing


Maxwell-Alexius:
As described from PR #470, float parsing cases below are normal:

```ruby
(3.14159265358979).to_d.to_s # => "3.14159265358979"
(-273.150000000).to_d.to_s # => "-273.15"

# and without the round bracket
3.14159265358979.to_d.to_s # => "3.14159265358979"
```

However, there is a case when a negative valued float type number cannot call method without round bracket
```ruby
-273.150000000.to_d.to_s # ERROR
```

st0012:
This issue blocks #379

hachi8833:
I fixed the issue locally and would send the PR tomorrow. Wait a bit. 

Before that, I'd create the following issue to close together after that:

- current compiler contains unused `Incr     = "++"` and `Decr     = "--"`
- handling unary operators: `+1` and `-1`

---------------

Well, I found that Ruby treats `+++++1` as `1`, and `--1`  or `----1` as `1`, and `---1` as `-1`.

PR #557, opened 2017-12-31
[WIP] Fix minus literal

hachi8833:
To close https://github.com/goby-lang/goby/issues/379 and to fix:

* https://github.com/goby-lang/goby/issues/526
* https://github.com/goby-lang/goby/issues/554
* https://github.com/goby-lang/goby/issues/555

Currently `-1.class` or `-1.0.class` works, but I'm wondering how deep-duplicated operators like `--+- -++-- -` or `1 + + - + 1`.

* duplicated minus operator with spaces like `1- -  - 1` works.
* duplicated plus operator (or mixed) with spaces like `1+ ++ -- +1` does not work well.

I guess the implementation should be more simplified.

Looks like I'm catching cold and will be back soon.


hachi8833:
Updated the PR and fixed most issues. I believe the PR has been simplified than the initial one.

The remaining issues are related with `;`:

```ruby
» a=1; +a
#» 2 # invalid: should be 1
```

FYI: current master is also affected by the issue.

compiler/lexer/lexer.go
```diff
@@ -389,6 +399,69 @@ func (l *Lexer) peekChar() rune {
...
+	}
+}
+
+func (l *Lexer) skipUnary() {
+	if l.ch == '+' && l.isUnary() {
+		l.readChar()
+	}
+}
+
+func (l *Lexer) isUnary() bool {
```
st0012: I think this should be a helper function that receives an `rune` argument just like `isIdentifier`, instead of a `lexer`'s function. We should change either its behavior or its name for the consistency.

hachi8833: First I thought as that. But this helper function goes backward the lexer so `*Lexer` is needed.
Passing `*Lexer` via argument might be ineffective.
I will change the function name to `checkUnary()`.

hachi8833:
https://github.com/goby-lang/goby/issues/526 has been resolved by https://github.com/goby-lang/goby/pull/569.
Thus I close this PR and then I will create a new PR based on https://github.com/goby-lang/goby/pull/569.

Issue #403, opened 2017-09-11
Interpreter crashing on sliced Array access


saveriomiroddi:
When I access an array in sliced form `Array[start, length]`, the interpreter crashes.

Sample command:

    a = [0, 1, 2]
    a[0, 3] # kaboom!

Exact run output (the interpreter overwrites the second line when crashing):

    $ ./goby -i
    Goby 0.1.2 😙 😇 🤔
    » a = [0, 1, 2]
    #» [0, 1, 2]
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x7114c5]

    goroutine 1 [running]:
    github.com/goby-lang/goby/compiler/ast.(*BaseNode).MarkAsExp(0x0)
      /home/saverio/code/go/src/github.com/goby-lang/goby/compiler/ast/ast.go:36 +0x5
    github.com/goby-lang/goby/compiler/parser.(*Parser).parseStatement(0xc420114240, 0xfdb598, 0x1)
      /home/saverio/code/go/src/github.com/goby-lang/goby/compiler/parser/statement_parsing.go:34 +0xbd
    github.com/goby-lang/goby/compiler/parser.(*Parser).ParseProgram(0xc420114240, 0x7, 0xc420063e40)
      /home/saverio/code/go/src/github.com/goby-lang/goby/compiler/parser/parser.go:176 +0xdf
    github.com/goby-lang/goby/igb.StartIgb(0xb1be56, 0x5)
      /home/saverio/code/go/src/github.com/goby-lang/goby/igb/repl.go:160 +0x4d1
    main.main()
      /home/saverio/code/go/src/github.com/goby-lang/goby/goby.go:29 +0x954

===============

- Your operating system
Ubuntu 16.04 64bit
- How did you install Goby
From source
- Goby version
93d2e10fecedf5451e517afbcee2b8e275ad5217
- The value of `$GOBY_ROOT`
$HOME/code/go/src/github.com/goby-lang/goby

st0012:
@saveriomiroddi Can you help fixing this?

saveriomiroddi:
@st0012 I need a couple of days to get acquainted with the lexer (or parser) :smile: 

I will write back with a positive or negative answer in a couple of days :smile:

saveriomiroddi:
@st0012 ok, I had a look at the parser, and definitely manual parsers are not my thing :smile: so I'll pass on this, sorry.

[...]

PR #559, opened 2018-01-02
#403 Able to Index with Successive Values from Array

Maxwell-Alexius:
This PR should provide the utility for Goby to support and fulfill the requirement of the issue #403. The following example is implemented in this PR:

```
>> a = [1, 2, 3, 4, 5, 6, 7]
>> a[1, 3] # Start from index 1, get 3 values until meet the end
=> [2, 3, 4]
>> a[1, 8] # Start from index 1, get 8 values until meet the end
=> [2, 3, 4, 5, 6, 7]
>> a[-4, 3] # Start from index a.length - 4 = 3, get 3 values until meet the end
=> [4, 5, 6]
>> a[9, 1] # Start from index 9 => exceeded, return nil
=> nil
>> a[-10, 3] # Start from index a.length - 10 = -3, invalid starting index, return nil
```

~Currently I haven't rescue the case where~
```
a[1, 2, 3] # with 3 arguments will raise ArgumentError
```
[Solved]

~Second, I haven't implemented this yet but this should be include in this PR:~

```
>> a = [1, 2, 3, 4, 5, 6, 7]
>> a[1, 3] = [9, 8, 7]
=> [1, 9, 8, 7, 5, 6, 7]
```
[Solved and Implemented]

~Lastly, it should also ban the condition when trying to do something like this (I think so?):~

```
>> a = { hello: "World", lang: "Goby" }
>> a[1, 2]

or

>> a['hello', 'world'] # Should raise error
```
[Solved]


Maxwell-Alexius:
[Update] Well, as I inspect Ruby's behaviour, I found out that actually the `[]` method can accepts multiple arguments like this:

```Ruby
2.3.3 :001 > a = [1,2,3,4,5]
 => [1, 2, 3, 4, 5] 
2.3.3 :002 > a[1, 4, 6]
ArgumentError: wrong number of arguments (given 3, expected 1..2)
        from (irb):2:in `[]'
        from (irb):2
        from /Users/MaxwellAlexius/.rvm/rubies/ruby-2.3.3/bin/irb:11:in `<main>'
2.3.3 :003 > 
```

So I decided to let Goby support this behaviour which will become like this also:
```
╰─⠠⠵ go run goby.go -i
# command-line-arguments
ld: warning: PIE disabled. Absolute addressing (perhaps -mdynamic-no-pic) not allowed in code signed PIE, but used in type..eqfunc.[106]string from /var/folders/70/jfl6pkgn39l1xxbg70r4cqdc0000gn/T/go-link-088606754/go.o. To fix this warning, don't compile with -mdynamic-no-pic or link with -Wl,-no_pie
Goby 0.1.6 😂 🤠 😳
» a = [1,2,3,4,5]
#» [1, 2, 3, 4, 5]
» a[1, 3]
#» [2, 3, 4]
» a[1, 3, 5]
#» ArgumentError: Expect 1..2 arguments. got=3
» a[1, 3, 5, 7, 8]
#» ArgumentError: Expect 1..2 arguments. got=5
» b = { hello: 'Taipei', location: 'Taiwan' }
#» { hello: "Taipei", location: "Taiwan" }
» b["hello"]
#» Taipei
» b["hello", "location"]
#» ArgumentError: Expect 1 argument. got: 2
»  
```

Which means the `[]=` can also support multiple arguments as well (that is so sexy, I thought)

compiler/parser/expression_parsing.go
```diff
@@ -241,9 +241,15 @@ func (p *Parser) parseIndexExpression(left ast.Expression) ast.Expression {
 	}
 
 	p.nextToken()
-
 	callExpression.Arguments = []ast.Expression{p.parseExpression(NORMAL)}
 
+	// Accepting multiple indexing argument
+	for p.peekTokenIs(token.Comma) {
```
Maxwell-Alexius: This enables us to perform something like this:

```
>> a = [1, 2, 3, 4, 5]
[1, 2, 3, 4, 5]
>> a[1, 2, 3]
ArgumentError: Expect 1..2 arguments. got=3
```

Maxwell-Alexius:
Hey~ @hachi8833, would you want to review this, this might take a little time to review because it got a TONS of cases (see test case in File Changes) @@

vm/array_test.go
```diff
@@ -140,16 +140,214 @@ func TestArrayIndex(t *testing.T) {
...
+}
+
+func TestArrayIndexWithSuccessiveValuesNullCases(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		{`
+			a = [1, 2, 3, 4, 5]
+			a[5, 5] # Range exceeded
```
hanachin: I check the ruby/spec to confirm the ruby's behavior, in this case, ruby returns `[]`.
https://github.com/ruby/spec/blob/875da49e051c030db05cc1a6bcd4a440e9266389/core/array/shared/slice.rb#L52-L53

```
% ruby -ve 'a = [1, 2, 3, 4, 5]; p a[5, 5]'
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-linux]
[]
```

vm/array.go
```diff
@@ -130,26 +130,92 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
...
+					}
+
+					// <Two Argument Case>
+					// Second argument is the assignment value
+
 					// Negative index value condition
 					if indexValue < 0 {
 						if len(arr.Elements) < -indexValue {
-							return t.vm.initErrorObject(errors.ArgumentError, sourceLine, "Index is too small for array. got=%s", i.Class().Name)
+							return t.vm.initErrorObject(errors.InternalError, sourceLine, "Index value %d too small for array. minimum: %d", indexValue, -arr.length())
```
st0012: I should `index value too small` should still be `ArgumentError`?

vm/array_test.go
```diff
@@ -140,16 +140,214 @@ func TestArrayIndex(t *testing.T) {
...
+			a[5, 123] = 123
+			a
+		`, []interface{}{1, 2, 3, 4, 5, 123}},
+		{`
+			a = [1, 2, 3, 4, 5]
+			a[5, 123] = [1, 2, 3]
+			a
+		`, []interface{}{1, 2, 3, 4, 5, 1, 2, 3}},
+		{`
+			a = [1, 2, 3, 4, 5]
```
st0012: We should document this kind of cases, can you add some comments (or references) here to explain that?

Maxwell-Alexius: Alright, I shall document these behaviour in these days, took me a bit time to clarify these bizarre cases

vm/array.go
Maxwell-Alexius: I'll change it to `ArgumentError` ~

vm/array_test.go
Maxwell-Alexius: Thanks a lot for reminding this special case! <3  i'll change it later on!

hachi8833: Double negative can be checked: 

```ruby
a = [1,2,3,4,5]
» a[-1,-1]
#» nil
» a[-1,-1] = 555
#» InternalError: Expect second argument greater than or equal 0. got: -1
```

Zero case:

```ruby
a = [1,2,3,4,5]
#» [1, 2, 3, 4, 5]
» a[1,0]
#» []
» a[1,0] = 555
#» 555
» a
#» [1, 555, 2, 3, 4, 5]

**Comment by Maxwell-Alexius** [comment]
[Update]

- Adding cases addressed by @hanachin && @hachi8833 
- Adding doc for `Array#[]` and `Array#[]=`  (See `vm/array.go`)
- Adding one more case which is discovered accidentally (originally it didn't raise error...):

```
a = [1, 2, 3, 4, 5]
a[6, -4] = [9, 8, 7] # <--- Should raise error
# ArgumentError: Expect second argument greater than or equal 0. got: -4
```

File: vm/array_test.go (hunk 4 above)
Comment (Maxwell-Alexius) [reply vm/array_test.go]: I added to the `vm/array.go` part

File: vm/array.go [hunk 5]
```diff
@@ -46,6 +46,32 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
...
+			// a[5, 1]  # => ["c"]
+			// a[6, 1]  # => []
+			// a[7, 1]  # => nil
+			//
+			// Special case 1:
+			// a[6]    # => nil
+			// a[6, 1] # => []  <-- Not nil!
+			// a[7, 1] # => nil <-- Because it is really out of the edge of the array
+			//
+			// Special case 2: Second argument is negative
```
st0012: According to the tests, I thought the second argument can't be negative?

Maxwell-Alexius: According to `irb`:
```
╰─⠠⠵ irb
2.4.1 :001 > a = [1,2,3,4,5]
 => [1, 2, 3, 4, 5] 
2.4.1 :002 > a[1, -1]
 => nil 
2.4.1 :003 > a[6, -1]
 => nil 
2.4.1 :004 > a[-3, -1]
 => nil 
2.4.1 :005 > a[-8, -1]
 => nil 
2.4.1 :006 > 
```

Maxwell-Alexius: In Goby, should we always raise error when encounter the second argument negative case?

st0012: I think we should. Will the result always be `nil` when the second argument is negative?

Maxwell-Alexius: Okay, let's change the behaviour when inputing the negative value in second argument! => It should raise `ArgumentError` instead of `nil`

vm/array.go
```diff
@@ -125,31 +151,142 @@ func builtinArrayInstanceMethods() []*BuiltinMethodObject {
...
+			// a[-5, 4] = [9, 8, 7] # <-- Negative index edgy case
+			// a # => [9, 8, 7, 5]
+			//
+			// a = [1, 2, 3, 4, 5]
+			// a[-6, 4] = [9, 8, 7] # <-- Nagative index too small case
+			// # ArgumentError: Index value -6 too small for array. minimum: -5
+			//
+			// # While indexing, it returns nil instead error
+			// a[-6, 4] # => nil
+			//
```
Maxwell-Alexius: Hi, all @goby-lang/maintainers, should we also change this to this kind of behaviour?

```
a = [1, 2, 3, 4, 5];
a[-6, 4] # => ArgumentError: Index value -6 too small for array. minimum: -5
```

Because in Ruby, originally:
```ruby
a = [1, 2, 3, 4, 5];
a[-6, 4] # => nil
```

It returns `nil` instead of errors

st0012: Should raises error I think

saveriomiroddi: This is a very tricky subject.

Ruby consistently never raise an error on Array indexed access (reading); in fact, exceptions are raised only on invalid Array writing.

If it's decided that the out of bound selection on the "left" (negative) side will raise an exception, also the mirrored "right" (positive) side should:

    [0, 1, 2][-3, 2] # => error
    [0, 1, 2][3, 2] #=> error

Ultimately, before taking the decision, it should be clarified very precisely what's the rationale behind Ruby never raising an error on Array access (read), and evaluate if that principle should consisently apply.

Maxwell-Alexius: [Update]
I've pushed a commit in this PR - when indexing array, even if it has only one parameter, if the index value is negatively too small, it will raise the `ArgumentError` no matter what, the same applied to the `ConcurrentArray`:

```
a = [1, 2, 3, 4, 5]

a[-6]
# => ArgumentError: Index value -6 too small for array. minimum: -5
a[-6] = 123
# => ArgumentError: Index value -6 too small for array. minimum: -5
a[-6, 1]
# => ArgumentError: Index value -6 too small for array. minimum: -5
a[-6, 1] = 123
# => ArgumentError: Index value -6 too small for array. minimum: -5

require 'concurrent/array'
ca = Concurrent::Array.new([1, 2, 3, 4, 5])

ca[-6]
# => ArgumentError: Index value -6 too small for array. minimum: -5
ca[-6] = 123
# => ArgumentError: Index value -6 too small for array. minimum: -5
ca[-6, 1]
# => ArgumentError: Index value -6 too small for array. minimum: -5
ca[-6, 1] = 123
# => ArgumentError: Index value -6 too small for array. minimum: -5
```

PR #563, opened 2018-01-03
Fix #540

### LINKED ISSUE #540: Blocks don't return nil when empty
There is no current specification for this - I think it's a bug, but it may not necessarily be.

When an empty block is passed to a function, `nil` value is not returned; instead, I think that the caller itself is returned. For example

```ruby
[1, 2].map do end # => [[1, 2], [1, 2]]
[1, 2].select do end # => [1, 2]
```

I would expect `[nil, nil]` in the first case, and `[]` in the second, as I think `nil` is an appropriate value to return when a block is empty.

hachi8833:
- [x] Array
- [x] String
- [x] Hash
- [x] Integer
- [x] Class
- [x] Range
- [x] Concurrent_hash
~~- [ ] Concurrent_rw_lock~~

I will squash the commits when finished.


st0012:
@hachi8833 Is there any blocker for this PR? maybe I can help

Issue #537, opened 2017-12-12
Implement Enumerable#lazy


saveriomiroddi:
I'm adding this for tracking purposes, based on the Slack discussion.

I'm not sure I will have the time over the next couple of weeks, however, I'm posting here the reference to the commit introducing the feature in Ruby:

- https://github.com/ruby/ruby/commit/0b2c4f43d27510cd0f51a260b5b21fc7e39a5e85

EDIT: this is a much easier to read (and implement) version, from JRuby:

- https://github.com/jruby/jruby/blob/master/core/src/main/ruby/jruby/kernel/enumerator.rb
- https://github.com/jruby/jruby/blob/master/core/src/main/ruby/jruby/kernel/enumerable.rb

st0012:
@saveriomiroddi thanks for the reference, and it's not rush to implement this feature actually. I think we won't touch anything related to this until next January or even February.

saveriomiroddi:
So, the logic of the lazy enumerators itself is not so complicated (at least, in a basic/naive form).

There is a big problem, and it's that in Goby, blocks are not objects. so they can't be passed around and executed (this is AFAIK; things may have changed).

The below implementation is the simplest possible Ruby one (supporting `#each` and `#map`), and it's meant to be essentially copy/pasted to Goby once Goby has the required primitives.
The only changes to apply are the StopIteration convention, and the block invocation syntax.
A RangeIterator should be easy to create, but a way to store a current position in RangeObject should be
found (at least, following this implementation's design).

```ruby
class LazyEnumerator
  def initialize(parent, &block)
    @parent = parent
    @enumerator_block = block
  end

  def each(&user_block)
    loop do
      user_block[next_value]
    end
  rescue StopIteration
  end

  def map(&user_block)
    LazyEnumerator.new(self, &user_block)
  end

  def next_value
    result = @parent.next_value
    result = @enumerator_block[result] if @enumerator_block
    result
  end

  def first(size)
    # We need not to perform a number of :each cycles exactly the number of times required.
    # If the size is 0, we can't enter the block and break immediately.
    #
    return [] if size == 0

    result = []

    each do |value|
      result << value
      break if result.size >= size
    end

    result
  end
end

class ArrayEnumerator
  def initialize(array)
    @array = array
    @current_position = -1
  end

  def next_value
    @current_position += 1

    if @current_position >= @array.size
      raise "No next value!"
    else
      @array[@current_position]
    end
  end
end

class Array
  def lazy
[...]

PR #566, opened 2018-01-06
Implement #546 (Block object)

### LINKED ISSUE #546: Support block object
Currently, it's not possible to pass around (and execute) block object. This is a very useful feature

st0012:
This closes #546


vm/evaluation_test.go
```diff
@@ -1509,6 +1509,44 @@ func TestMethodCallWithBlockArgument(t *testing.T) {
 			  ten + 20
 			end
 			`, 30},
+		// Get Block
+		{`
```
st0012: @saveriomiroddi Are these what you need to implement #537?

vm/block_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import "testing"
+
+func TestBlockInitialize(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected interface{}
+	}{
+		{`
+b = Block.new do
```
st0012: I also implemented something like `Proc.new`. But I always don't think `Proc` is a good name, so I just call it `Block`.
And I **won't** support something like `proc`.

vm/block_test.go
```diff
@@ -0,0 +1,49 @@
...
+  100
+end
+
+b.call
+`, 100},
+
+		{`
```
st0012: @saveriomiroddi @hachi8833 Do you think we need more complex cases like this?

hachi8833: Looks sufficient by now.

st0012:
@saveriomiroddi We already have `block_given?` for a while, but it exists as a method instead a keyword.
I'll add more test cases for both `get_block` and `block_given?`.

vm/evaluation_test.go
saveriomiroddi: > Are these what you need to implement #537?

@st0012 yes, I think so!

PR #573, opened 2018-01-12
Lazy Enumerator, with lazy array enumeration

saveriomiroddi:
Basic implementation of a lazy enumerator, with Array support for it via Array#lazy.

Requires #572, ~#566~, #578.

### DIFF
--- a/lib/array.gb
+++ b/lib/array.gb
@@ -8,4 +8,10 @@ class Array
   def to_enum
     ArrayEnumerator.new(self)
   end
+
+  # Return a lazy iterator for self.
+  #
+  def lazy
+    LazyEnumerator.new(to_enum)
+  end
 end
--- a/lib/lazy_enumerator.gb
+++ b/lib/lazy_enumerator.gb
@@ -1,0 +1,96 @@
+# Wrapper class for implementing enumerators, whose blocks are not executed
+# unless required.
+# This is useful for sets that can't be fully enumerated (eg. because they're
+# too slow), and that are typically only partially enumerated and then halted.
+#
+# Chaining is supported, for the methods `#each` and `#map`.
+#
+# Basic example:
+#
+#   [0, 1, 2].lazy.map { |i| puts "Iterated: #{i}"; i }.first(2)
+#   Iterated: 0
+#   Iterated: 1
+#   => [0, 1]
+#
+# Chaining example:
+#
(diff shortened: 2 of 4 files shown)


st0012:
@saveriomiroddi #566 is merged now!

st0012:
@saveriomiroddi I think all blocking PR have been merged now

saveriomiroddi:
@st0012 ready :smile:

st0012:
@saveriomiroddi LGTM

Will our next step be supporting lazy enumerator on range? I think it'll be important as well and can show the power of lazy evaluation more.

```ruby
module Enumerable
  def filter_map(&block)
    map(&block).compact
  end
end

class Enumerator::Lazy
  def filter_map
    Lazy.new(self) do |yielder, *values|
      result = yield *values
      yielder << result if result
    end
  end
end

(1..10000).lazy.filter_map{|i| i*i if i.even?}.first(5)
    # => [4, 16, 36, 64, 100]
```

(I know the code will look differently, but I think we should be able to achieve this, right?

saveriomiroddi:
> Will our next step be supporting lazy enumerator on range? I think it'll be important as well and can show the power of lazy evaluation more.

Yes. The conventional showcase for lazy evaluation uses `Float::INFINITY`:

```ruby
(1..Float::INFINITY).lazy.map { |i| 2*i }.first(5) # => [2, 4, 6, 8, 10]
```

Because examples of the type:

```ruby
(1..10000).lazy.map { |i| i*i if i.even? }.first(5)       # example via :map, as Ruby has no :filter_map
```

don't give an obvious difference in an example run; using Float::INFINITY will hang without lazy iterator.

saveriomiroddi:
In order to implement lazy range enumeration, a RangeEnumerator will need to implemented, which has the methods `#next` and `#has_next?`.

st0012:
@saveriomiroddi So if there's any good showcase for Array's lazy evaluation?

saveriomiroddi:
> So if there's any good showcase for Array's lazy evaluation?

Not really, in fact, when I showcased or tested, I used either a conditional error raise, or an accumulation array (with check). Possibly the conditional error raise is the best:

    [0, 1, 2].lazy.map { |n| n >= 2 ? raise : n }.first(2)

In general, lazy evaluation is always shown with a range (and infinity), however this also shows clearly lazy evaluation (although it's somewhat unconventional).

PR #579, opened 2018-01-16
Allow calling break inside a block.

st0012:
Todo this, we need different mechanism than while statement's break.
Cause if we want to break from a block, it means we want to leave
current frame and scope.

So the solution is to pop every callframe related to that block and stop
their execution.

### DIFF
--- a/compiler/bytecode/instruction.go
+++ b/compiler/bytecode/instruction.go
@@ -36,6 +36,7 @@ const (
 	BranchUnless        = "branchunless"
 	BranchIf            = "branchif"
 	Jump                = "jump"
+	Break               = "break"
 	DefMethod           = "def_method"
 	DefSingletonMethod  = "def_singleton_method"
 	DefClass            = "def_class"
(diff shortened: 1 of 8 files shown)


st0012:
I want to discuss the following behavior:

```
[1] pry(main)> [1, 2].map { break }
=> nil
[2] pry(main)> [1, 2].map { |i|  break if i == 2 } 
=> nil
[3] pry(main)> [1, 2].map { |i|  break if i == 1 }
=> nil
[4] pry(main)> [1, 2].map { |i|  break if i == 0 }
=> [nil, nil]
```

I think it's quite inconsistent. I prefer the `map` always returns an array instead. So it'll be

```
[1] pry(main)> [1, 2].map { break }
=> [nil]
[2] pry(main)> [1, 2].map { |i|  break if i == 2 } 
=> [nil, nil]
[3] pry(main)> [1, 2].map { |i|  break if i == 1 }
=> [nil]
[4] pry(main)> [1, 2].map { |i|  break if i == 0 }
=> [nil, nil]
```

What do you think? @saveriomiroddi @hachi8833

hachi8833:
>  I prefer the map always returns an array instead.
I agree with that.

PR #583, opened 2018-01-22
Minimal Range enumerator implementation

saveriomiroddi:
Minimal implementation of a range enumerator; this allows lazy enumeration via LazyEnumerator.

### DIFF
--- a/lib/range.gb
+++ b/lib/range.gb
@@ -1,0 +1,9 @@
+class Range
+  def lazy
+    LazyEnumerator.new(to_enum)
+  end
+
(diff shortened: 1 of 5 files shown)


saveriomiroddi:
Ready :smile:

st0012:
@saveriomiroddi So if I finish the `Float::Infinite` then we can have the classic lazy evaluation showcase?

saveriomiroddi:
> So if I finish the Float::Infinite then we can have the classic lazy evaluation showcase?

Yes.

PR #587, opened 2018-02-03
Implement #550

### LINKED ISSUE #550: Better way to test AST
Currently parser tests are very inefficient cause we need to assert every node like

```go
stmt := program.Statements[0].(*ast.ClassStatement)
```

And I want to have some pattern like:

```go
input := `
	class Foo
	  def bar(x, y)
	    x + y
	  end
	end
	`

l := lexer.New(input)
p := New(l)
program, err := p.ParseProgram()

if err != nil {
	t.Fatal(err.Message)
}

// Actually we should pass `t` into `hasClassStmt`
// but I want to avoid passing `t` to every verify function
classStmt := program.firstStmt().hasClassStmt("Foo")
methodStmt := classStmt.hasMethodStmt("bar")
methodStmt.hasParams("x", "y")
```

I think this is more readable and easy to write, what do you think? @saveriomiroddi @hachi8833

st0012:
This closes #550 

I've implemented some functions to make this work. You can see `compiler/parser/statement_parsing_test.go` below as an usage example.

~~But this creates another issue, which is we are coupling our test helper functions with our AST nodes. This will make the compiled binary contains some useless functions. And I don't think we can get rid of them by sending options to the compiler, because we also touched the interface for those functions.~~

And in order to avoid compiling those test helper function/types, we need to add `-tags release` when compiling the program.


st0012:
To do that, we’ll also need to be able to change the interfaces according to the tags. I’m thinking maybe we can use embedded type to create test-only interfaces

saveriomiroddi:
> To do that, we’ll also need to be able to change the interfaces according to the tags.

That can't be done in Go :cry: The two different interfaces will need to be defined in two separate files; this is, though, the purpose of conditional compilation is for :smile:

saveriomiroddi:
:+1:

st0012:
@Sir Diva Moodier Currently this branch has build issue. But I can’t reproduce it on my machine (macOS). Also, the makefile works fine in both macOS and linux (docker). Only `travis.sh` will fail after the VM's test in linux environment. Can you help me investigate the cause?

saveriomiroddi:
This is a log of the commit travis runs the build against:

```
$ git log
commit 3172f2cd441a932c88450858d5c532713d4d59d2 (HEAD -> failure)
Merge: 87fcab8 fc96816
Author: Stan Lo <stan001212@gmail.com>
Date:   Sun Feb 4 15:57:47 2018 +0000

    Merge fc968162dd165d1886127bd5ae993cf4b5133c5b into 87fcab8eb77d1ca2065de5b2c7b3aa8c4d74b544

commit fc968162dd165d1886127bd5ae993cf4b5133c5b
Author: st0012 <a22301613@yahoo.com.tw>
Date:   Sun Feb 4 23:32:30 2018 +0800

    Use build flag to avoid compiling test helper types/functions.
```

st0012:
@saveriomiroddi Got it, thanks a lot!

st0012:
@goby-lang/maintainers Can you help me review the new test pattern in `compiler/parser`? I think I made it more semantically friendly but need some feedback.

st0012:
I'll merge this first and add comments when I have time

PR #600, opened 2018-02-21
Support string range indexing

st0012:


st0012:
@hachi8833 @saveriomiroddi This should solve the floating point comparison issue.

PR #603, opened 2018-02-25
Fix #444

### LINKED ISSUE #444: Arithmetic operations with multiple right values are allowed
In Goby, it's allowed to perform an arithmetic operation with two right values:

    1 * *[2, 3] #=> 2

I think this should be forbidden; if so, probably the best place to forbid this is at parser level, otherwise, argument checks should be added for each operation method.

ear7h:
This fixes the issue.

I did not include tests since I couldn't find a proper place for it. It makes sense that they wouldn't exist, 
 testing for every syntax error would get out of hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,8 @@ See the [guideline](https://github.com/goby-lang/goby/blob/master/CONTRIBUTING.m
 
 - @st0012
 - @hachi8833
-- @Maxwell-Alexius
 - @saveriomiroddi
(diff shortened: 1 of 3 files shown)


st0012:
Can you use `TestHashExpressionFail` as an example to write at least two test cases to prevent regression?

ear7h:
@st0012 Okay, I will do. Thanks for giving a direction.

Issue #602, opened 2018-02-24
Add integration tests for simple server


st0012:
Since simple server is one of our big features, it should be covered with integration tests to prevent regression

PR #604, opened 2018-02-25
Implement #602 and fix simple server

st0012:
Previously because of implementing `break` I broke our nested block execution, which caused simple server's block argument unable to function normally. So this PR is to add integration test for simple server and fix the issue.

### DIFF
--- a/vm/simple_server.go
+++ b/vm/simple_server.go
@@ -13,6 +13,8 @@ import (
 	"github.com/fatih/structs"
 	"github.com/goby-lang/goby/vm/classes"
 	"github.com/gorilla/mux"
+	"fmt"
+	"strconv"
 )
 
 type request struct {
@@ -71,12 +73,19 @@ func builtinSimpleServerInstanceMethods() []*BuiltinMethodObject {
 					var serveStatic bool
 					server := receiver.(*RObject)
 
-					portVar, ok := server.InstanceVariables.get("@port")
+					portVar, ok := server.instanceVariableGet("@port")
 
 					if !ok {
 						port = "8080"
 					} else {
-						port = portVar.(*StringObject).value
+						switch p := portVar.(type) {
+						case *StringObject:
+							port = p.value
+						case *IntegerObject:
+							port = strconv.Itoa(p.value)
+						default:
+							fmt.Printf("Unexpected type %s for port setting\n", portVar.Class().Name)
+						}
 					}
 
 					log.Println("SimpleServer start listening on port: " + port)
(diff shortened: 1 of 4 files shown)


vm/simple_server_test.go
```diff
@@ -28,6 +31,53 @@ func TestServerInitialization(t *testing.T) {
...
+			200},
+	}
+
+	for _, tt := range tests {
+		go func() {
+			v := initTestVM()
+			v.testEval(t, tt.input, getFilename())
+		}()
+
+		time.Sleep(1 * time.Second)
```
ear7h: This is my only concern. Would running `server.start` from a thread cause execution to end in the test VM? If not, you could use a mutex.

st0012: What do you mean? It'll make the vm inside goroutine to run the server (and hang there). This sleep is just to make sure we don't make the request before the server is started

ear7h: I metioned the wrong syncing mechanism, meant a [wait group](https://golang.org/pkg/sync/#WaitGroup). I just think we shouldn't rely on `time.sleep` for thread safety. I was curious if the following snippets would work:

```ruby
thread do
   server.start
end
```
the tests function 
```go
waitgroup.Add(1)
go func() {
    ...
    v.testEval(...) // does the vm die after this statement? or will the server still be running?
    waitgroup.Done()
}

waitgroup.Wait()
// run the tests

Comment (st0012) [reply vm/simple_server_test.go]: @ear7h The problem is that since we're running a web server in `testEval`, so it won't actually end and call `waitgroup.Done()`

Comment (ear7h) [reply vm/simple_server_test.go]: Ah okay, I wasn't sure of the mechanics of the test vm

File: vm/simple_server.go [hunk 2]
```diff
@@ -10,10 +10,10 @@ import (
 	"path/filepath"
 	"unicode"
 
+	"fmt"
```
ear7h: should this be with the group above?

vm/simple_server.go
```diff
@@ -10,10 +10,10 @@ import (
...
 	"github.com/fatih/structs"
 	"github.com/goby-lang/goby/vm/classes"
 	"github.com/gorilla/mux"
-	"fmt"
 	"strconv"
```
ear7h: also grouped with the std lib

st0012: @ear7h I already ran `gofmt` and I want to follow what it changes/allows

ear7h: Yes, but I believe gofmt preserves newlines so the two groups of imports are formatted independently of each other.

st0012: If you insist I can change this time, but I don't see the point to group them all manually in every future PRs

ear7h: Alright, makes sense

PR #615, opened 2018-03-07
[WIP] Adding VM Float API

### LINKED ISSUE #484: VM Float API Implementation
Referenced from the Ruby Ver. 2.4.1. Float data type API:

- [ ] Constant `Float::INFINITE` stands for the number is infinite
- [ ] Constant `Float::NAN` stands for "Not a Number"

- [ ] `Float#modulo` Synonym to the `%` operator (Get the quotient from the divide operation)
- [ ] `Float#quo` Synonym to the `/` operator (Get the quotient from the divide operation)
- [ ] `Float#abs` Return positive number of the float. "Formally as the distance between 0 to the number of the float in 1-dimensional axis ..."
- [ ] `Float#angle` (and `Float#arg`) If positive -> return 0; else -> return PI
- [ ] `Float#ceil` Returns smallest number greater than or equal to the receiver
- [ ] `Float#floor` Returns largest number smaller than or equal to the receiver
- [ ] `Float#round` Rounds float to a given precision in decimal digits (default 0 digits). Precision may be negative. Returns a floating point number when `n` digits is more than zero.
- [ ] `Float#finite?` Returns `true` if receiver isn't  equal to `Float::INFINITE`
- [ ] `Float#infinite?` Returns `true` if receiver is equal to `Float::INFINITE`
- [ ] `Float#divmod` Returns array which is in format of [ (Integer) Quotient, (Float) Remainder ]
- [ ] `Float#nan?` Returns `true` if receiver is equal to `Float::NAN`
- [ ] `Float#positive?` Returns `true` if float number is larger than `0.0`
- [ ] `Float#negative?` Returns `true` if float number is less than `0.0`
- [ ] `Float#zero?` Returns `true` if float number is equal to `0.0`
- [ ] `Float#rationalize` Convert to something called `Rational` data type (Currently not supported but from my point of view, it is worth being addressed in this issue)

SD10:
Hey everyone 👋,

I started working on #484. However, I'm new to Goby, Go, and Ruby 😂 

This is classified as work in progress because I wanted to make sure I'm doing things correctly before trying to implement the full API. Any feedback is appreciated!

Referenced from the Ruby Ver. 2.4.1. Float data type API:
- [ ] Constant `Float::INFINITE` stands for the number is infinite

- [x] `Float#abs` Return positive number of the float. "Formally as the distance between 0 to the number of the float in 1-dimensional axis ..."
- [ ] `Float#angle` (and `Float#arg`) If positive -> return 0; else -> return PI
- [x] `Float#ceil` Returns smallest number greater than or equal to the receiver
- [x] `Float#floor` Returns largest number smaller than or equal to the receiver
- [ ] `Float#round` Rounds float to a given precision in decimal digits (default 0 digits). Precision may be negative. Returns a floating point number when `n` digits is more than zero.
- [ ] `Float#finite?` Returns `true` if receiver isn't  equal to `Float::INFINITE`
- [ ] `Float#infinite?` Returns `true` if receiver is equal to `Float::INFINITE`
- [ ] `Float#divmod` Returns array which is in format of [ (Integer) Quotient, (Float) Remainder ]
- [x] `Float#positive?` Returns `true` if float number is larger than `0.0`
- [x] `Float#negative?` Returns `true` if float number is less than `0.0`
- [x] `Float#zero?` Returns `true` if float number is equal to `0.0`

Lower priority:
- [ ] Constant `Float::NAN` stands for "Not a Number"
- [ ] `Float#nan?` Returns `true` if receiver is equal to `Float::NAN`
- [ ] `Float#rationalize` Convert to something called `Rational` data type (Currently not supported but from my point of view, it is worth being addressed in this issue)

This closes #484


st0012:
@SD10 for test cases, can you include Ruby doc's examples? Like https://ruby-doc.org/core-2.4.0/Float.html#method-i-ceil

SD10:
I'm implementing `divmod` now. All the others require `NAN` and `INFINITE` to be defined. @st0012 How do you think we should implement these constants? We also need `PI`

saveriomiroddi:
> I'm implementing divmod now. All the others require NAN and INFINITE to be defined.
> @st0012 How do you think we should implement these constants? We also need PI

I had a look at the IEEE754 specification, the MRI source code, and the Golang API, and there are a few interesting points:

- Positive/negative infinity [map to a single value each](https://en.wikipedia.org/wiki/IEEE_754-1985#Positive_and_negative_infinity), different only in the sign. For this reason, Ruby has only a single Float::INFINITY constant, the positive infinity (the sign can be used in the language grammar, which matches the binary representation)
- NaN is different, as there is [no single value representing it](https://en.wikipedia.org/wiki/IEEE_754-1985#NaN). I'm somewhat puzzled by why Ruby defines a constant for it - notice that the documentation ([Float::NAN [is] An expression representing a value which is “not a number”](https://ruby-doc.org/core-2.5.0/Float.html)) in fact says _a_ value, not _the_ value. Specifically, I'm puzzled why having a constant, since based on the reasons above, it can't be used for comparisons.

To summarize, based on my understanding, I think that:

- at least one Infinity constant, the positive one, is required; Go has APIs for getting the value (`math.Inf`);
- the Ruby `Float#infinite?` method is important;
- the `Float#nan?` method is required;
- I would leave the Float::NAN constant aside for now (because of `Float::NAN != Float::NAN`), waiting for somebody opening an issue with a use case.

SD10:
> - the Ruby Float#infinite? method is important;
 > - the Float#nan? method is required;

@saveriomiroddi Thanks for writing that up. I totally overlooked that we can still implement those methods.

st0012:
@saveriomiraddi thanks for the information, learned a lot 🙇‍♂️

eliothedeman:
`Float#nan?` is pretty trivial to impliment.

```ruby
a = Float::NAN
return a != a
```

NAN is the only float value for which this will be true.

st0012:
@SD10 Is it ok if I take over this PR?

SD10:
@st0012 Yes, no problems here 😓 Sorry

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

st0012:
@SD10  Since we have #664 for continuing the task, I'll close this

PR #618, opened 2018-03-09
Implement instance_eval

st0012:
Because we'll need this method for implementing test framework, so I added it. But it only accepts block argument.

If anyone comes up with any new test case, plz leave it in the comment

### DIFF
--- a/vm/class_test.go
+++ b/vm/class_test.go
@@ -1325,6 +1325,19 @@ func TestInstanceEvalMethod(t *testing.T) {
 `, "gnirtS"},
 		{`"a".instance_eval`, "a"},
 		{`"a".instance_eval do end`, "a"},
+		{`
+		class Foo
+		  def bar
+			10
+		  end
+		end
+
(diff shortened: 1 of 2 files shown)


vm/class_test.go
```diff
@@ -1297,6 +1297,73 @@ func TestClassSingletonClassMethod(t *testing.T) {
...
+		evaluated := v.testEval(t, tt.input, getFilename())
+		verifyExpected(t, i, evaluated, tt.expected)
+		v.checkCFP(t, i, 0)
+		v.checkSP(t, i, 1)
+	}
+}
+
+func TestInstanceEvalMethodFail(t *testing.T) {
+	testsFail := []errorTestCase{
+		{`"s".instance_eval(1, 1)`, "ArgumentError: Expect at most 1 arguments. got: 2", 1},
```
st0012: @saveriomiroddi here's the argument check, thanks for reminding 😄

PR #623, opened 2018-03-11
feature/issue-574

shes50103:
detect unexpected parsing error in parseDefMethodStatement() function.

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -218,3 +218,8 @@ func (p *Parser) callConstantError(t token.Type) {
 	msg := fmt.Sprintf("cannot call %s with %s. Line: %d", t, p.peekToken.Type, p.peekToken.Line)
 	p.error = errors.InitError(msg, errors.UnexpectedTokenError)
 }
+
+func (p *Parser) IsNotDefMethodToken() bool {
+
+	return p.curToken.Type != token.Ident && !(p.peekToken.Type == token.Dot && (p.curToken.Type == token.InstanceVariable || p.curToken.Type == token.Constant || p.curToken.Type == token.Self))
+}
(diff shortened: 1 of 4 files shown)


shes50103:
If there are any other parsing error which will crash parser please let me know

compiler/parser/statement_parsing.go
```diff
@@ -51,6 +51,11 @@ func (p *Parser) parseDefMethodStatement() *ast.DefStatement {
 
 	p.nextToken()
 
+	if !p.curTokenIs(token.Ident) && !(p.peekTokenIs(token.Dot) && (p.curTokenIs(token.InstanceVariable) || p.curTokenIs(token.Constant) || p.curTokenIs(token.Self))) {
```
st0012: I think you should refactor this to like `!p.IsArgumentToken()`

st0012:
@shes50103 I think we really need the recover mechanism, it's impossible to capture all invalid syntax cases.

shes50103:
@st0012 
okay I will add the recover mechanism. I am thinking that if there is a panic inside parser without error detected, despite panic is recovered in parser but the error statement still make repl.go crash.
I still need to set the error mechanism in most of function to prevent error from edge case.

compiler/parser/statement_parsing.go
shes50103: thanks, I will refactor it

compiler/parser/parser.go
```diff
@@ -112,6 +112,13 @@ func New(l *lexer.Lexer) *Parser {
 
 // ParseProgram update program statements and return program
 func (p *Parser) ParseProgram() (*ast.Program, *errors.Error) {
+
+	defer func() {
+		if recover() != nil {
+			fmt.Println("Panic has been recover parser ")
```
st0012: This can be improved to `Parser recovered from panic`

shes50103: thanks!

Issue #624, opened 2018-03-12
Assignment surrounded by parenthesis, inside a method, crashes the interpreter


saveriomiroddi:
An expression of the form:

```ruby
def test
  (@instance = 1)
end
```

crashes the interpreter.

If the parenthesis are removed, or the code is expressed in a single line, the interpreter won't crash.

This is the crash message/stack:

```
panic: interface conversion: ast.Expression is *ast.InstanceVariable, not *ast.Identifier

goroutine 1 [running]:
github.com/goby-lang/goby/compiler/bytecode.(*Generator).compileDefStmt(0xc4201d2d80, 0xc4201d6910, 0xc4201d9bc0, 0xc420177878)
  /path/to/workspace/go/src/github.com/goby-lang/goby/compiler/bytecode/statement_generation.go:203 +0x12fe
github.com/goby-lang/goby/compiler/bytecode.(*Generator).compileStatement(0xc4201d2d80, 0xc4201d6910, 0x1091d80, 0xc4201d9bc0, 0xc420177878, 0xc4201bfb00)
  /path/to/workspace/go/src/github.com/goby-lang/goby/compiler/bytecode/statement_generation.go:47 +0x50f
github.com/goby-lang/goby/compiler/bytecode.(*Generator).compileCodeBlock(0xc4201d2d80, 0xc4201d6910, 0xc4201bf9c0, 0xc420177878, 0xc4201bfb00)
  /path/to/workspace/go/src/github.com/goby-lang/goby/compiler/bytecode/generator.go:73 +0x7e
github.com/goby-lang/goby/compiler/bytecode.(*Generator).compileClassStmt(0xc4201d2d80, 0xc4201d67d0, 0xc4201d9b80, 0xc4201d2e10, 0xc4201bf640)
  /path/to/workspace/go/src/github.com/goby-lang/goby/compiler/bytecode/statement_generation.go:149 +0x42e
github.com/goby-lang/goby/compiler/bytecode.(*Generator).compileStatement(0xc4201d2d80, 0xc4201d67d0, 0x1091cc0, 0xc4201d9b80, 0xc4201d2e10, 0xc4201bf640)
  /path/to/workspace/go/src/github.com/goby-lang/goby/compiler/bytecode/statement_generation.go:49 +0x10b
[...]

PR #633, opened 2018-03-17
Feature/issue 624

shes50103:
Fix issue-624
Assignment surrounded by parenthesis, inside a method, crashes the interpreter #624

1. Ensure parameter is not instance variable (it will crash interpreter)
2. Identify  weather parenthesis statement is parameter or normal statement
which is make
```ruby=
def foo(a)
end
````
is different form
```ruby=
def foo
  (a)
end
```
The above codes get the same result in current master branch

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -231,3 +231,9 @@ func (p *Parser) IsNotDefMethodToken() bool {
 
 	return p.curToken.Type != token.Ident && !(p.peekToken.Type == token.Dot && (p.curToken.Type == token.InstanceVariable || p.curToken.Type == token.Constant || p.curToken.Type == token.Self))
 }
+
+// IsNotParamsToken ensures correct parameters which means it is not InstanceVariable
+func (p *Parser) IsNotParamsToken() bool {
+
+	return p.curToken.Type == token.InstanceVariable
+}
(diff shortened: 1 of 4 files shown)


compiler/parser/parser.go
```diff
@@ -231,3 +231,9 @@ func (p *Parser) IsNotDefMethodToken() bool {
 
 	return p.curToken.Type != token.Ident && !(p.peekToken.Type == token.Dot && (p.curToken.Type == token.InstanceVariable || p.curToken.Type == token.Constant || p.curToken.Type == token.Self))
 }
+
+// IsNotParamsToken ensures correct parameters which means it is not InstanceVariable
+func (p *Parser) IsNotParamsToken() bool {
+
+	return p.curToken.Type == token.InstanceVariable
```
st0012: I don't think this is just for instance variable. This should be a table lookup like

```go
_, ok := invalidParams[p.curToken.Type]
return ok
```

And that table should at least contain instance variable and constant

compiler/parser/parser.go
```diff
@@ -231,3 +231,17 @@ func (p *Parser) IsNotDefMethodToken() bool {
 
 	return p.curToken.Type != token.Ident && !(p.peekToken.Type == token.Dot && (p.curToken.Type == token.InstanceVariable || p.curToken.Type == token.Constant || p.curToken.Type == token.Self))
 }
+
+// Token type InstanceVariable and Constant will trigger IsNotParamsToken()
+var invalidParams = map[token.Type]bool{
```
shes50103: I think the return value bool is useless but I don't know whether there is any better way to implement this mapping table

st0012: I think this is the only way for that

shes50103:
@st0012 , there are some waning in codebeat do I need to fix it?
If we can ignore it then just merge it~

st0012:
@shes50103 they are actually existed issue I think, you just triggered the check. Let's merge it!

PR #639, opened 2018-04-02
Fix #636

### LINKED ISSUE #636: Forbidden include/extend on classes
This is related #611

st0012:
This closes #636


st0012:
@hachi8833 can you help me review this? and if this ok plz approve it

saveriomiroddi:
I've created [an issue](https://github.com/goby-lang/goby/issues/640) for tracking the error (in case it's going to be addressed separately).

st0012:
@saveriomiroddi nice catch! And I think that's because the way we return/retrieve value at the wrong stack position.

PR #645, opened 2018-04-11
Fix panic When a library file is not found on startup

saveriomiroddi:
Just return when there is an error - the invoking code will handle the error.

### DIFF
--- a/vm/thread.go
+++ b/vm/thread.go
@@ -79,7 +79,7 @@ func (t *thread) execFile(fpath string) (err error) {
 	file, err := ioutil.ReadFile(fpath)
 
 	if err != nil {
-		t.vm.mainThread.pushErrorObject(errors.InternalError, -1, err.Error())
+		return
 	}
 
 	instructionSets, err := compiler.CompileToInstructions(string(file), parser.NormalMode)
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/goby-lang/goby/compiler/parser"
 	"github.com/goby-lang/goby/vm/errors"
 	"os"
+	"path/filepath"
 	"runtime"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


saveriomiroddi:
> is it possible to write a test case for this?

@st0012 since the libraries are hardcoded in the go code, I think the only way is a hack - rename a library, perform the test, then restore the library.
Also, I don't think that right now we have test APIs for interpreter panic (although I'm not sure).

st0012:
If you can give us a piece of Goby sample code that can reproduce this issue, than we can test it. Also, panic happens inside the VM should be captured if my memory is right

saveriomiroddi:
From the project path:

```
rm lib/array.gb
./goby -i
```

st0012:
@saveriomiroddi can you describe the scenario that you encounter this issue? just curious

saveriomiroddi:
> can you describe the scenario that you encounter this issue? just curious

@st0012 hahah :smile: part of the packaging process is to copy the library files. on the first testing release of the package, the files weren't copied, so on execution, the goby interpreter would crash :laughing:

saveriomiroddi:
Let me know if you have any guideline about this type of testing - it's somewhat unusual (I think the only way is moving a `.gb` file out, then moving it back after the unit test).

saveriomiroddi:
> Although I think the best way of fixing this is to add the missing return, and fix the error initializing issue

:+1: in this case though, I'm not familiar with the code involving bug (the null reference error). is it because the VM searches for a file where it thinks there is an error, but it doesn't receive anything (because the error happens inside the Go code)?

st0012:
@saveriomiroddi when a program being executed, there'll be at least one callframe on the callframe stack. Each callframe contains a lot information about the current execution like the scope or some pointers (see [here](https://github.com/goby-lang/goby/blob/master/vm/call_frame.go#L29)).

And since the error happens even before the VM gets fully initialized, it's not surprise that there won't be any callframe on the callframe stack. So we got `nil` when calling `t.callFrameStack.top()` and the program crashed.

st0012:
I think I found the best solution, just try

```go
	if err != nil {
		return
	}
```

saveriomiroddi:
@st0012 all set, ready to go (once the checks are completed).

PR #646, opened 2018-04-11
Make lib path customizable on build, using the env variable GOBY_LIBPATH

saveriomiroddi:
This is a typical pattern for packaged interpreters, so that each binary can have arbitrary library paths, decided by the packager.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 GOFMT ?= gofmt -s
 GOFILES := $(shell find . -name "*.go" -type f -not -path "./vendor/*")
-RELEASE_OPTIONS := -ldflags "-s -w -X github.com/goby-lang/goby/vm.DefaultLibPath=${GOBY_LIBPATH}"
+RELEASE_OPTIONS := -ldflags "-s -w -X github.com/goby-lang/goby/vm.DefaultLibPath=${GOBY_LIBPATH}" -tags release
 TEST_OPTIONS := -ldflags "-s -w"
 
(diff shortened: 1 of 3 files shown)


st0012:
@saveriomiroddi we have `GOBY_ROOT` to specify the built-in lib's location already, so I think `GOBY_LIBPATH` is not necessary

saveriomiroddi:
ok, closing this as I think that then, this level of granularity is not useful.

Makefile
```diff
@@ -1,6 +1,6 @@
 GOFMT ?= gofmt -s
 GOFILES := $(shell find . -name "*.go" -type f -not -path "./vendor/*")
-RELEASE_OPTIONS := -ldflags "-s -w" -tags release
+RELEASE_OPTIONS := -ldflags "-s -w -X github.com/goby-lang/goby/vm.DefaultLibPath=${GOBY_LIBPATH}"
```
st0012: I think we should keep the release tag, because that prevents some testing purpose helper functions get compiled.

saveriomiroddi: Definitely. That's a mistake :blush:

st0012:
@saveriomiroddi I now understand what this is for. And for future maintenance, can you explain the purpose (your comment) in a readme section? or it'd be better if you can write it in our official doc https://github.com/goby-lang/goby-lang.github.io

saveriomiroddi:
Certainly will! :+1:

Issue #647, opened 2018-04-14
Support Integer conversion to Decimal


ear7h:
I don't think it'll take more than implementing a `to_d` method for the integer class.
```ruby
3.to_d #=> 3
```
#451 for context
cc @hachi8833

hachi8833:
Let me checking in!

Issue #648, opened 2018-04-14
Support user defined precision of Decimal strings


ear7h:
Currently, `Decimal#to_s` is hardcoded to support 60 digits of precision.

We should do something like (ruby)[https://ruby-doc.org/stdlib-2.2.0/libdoc/bigdecimal/rdoc/BigDecimal.html#method-i-to_s] which allows the user to pass a format string to the `to_s` method.
#451 for context 
cc @hachi8833

PR #649, opened 2018-04-14
Add Integer#to_d

hachi8833:
Fixes https://github.com/goby-lang/goby/issues/647

### DIFF
--- a/vm/integer.go
+++ b/vm/integer.go
@@ -393,6 +393,9 @@ func builtinIntegerInstanceMethods() []*BuiltinMethodObject {
 			Name: "to_d",
 			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
 				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
+					if len(args) > 0 {
+						return t.vm.initErrorObject(errors.ArgumentError, sourceLine, "Expect 0 arguments. got: %d", len(args))
+					}
 					r := receiver.(*IntegerObject)
 					return t.vm.initDecimalObject(intToDecimal(r))
 				}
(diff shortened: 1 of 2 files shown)


vm/integer.go
```diff
@@ -383,6 +383,21 @@ func builtinIntegerInstanceMethods() []*BuiltinMethodObject {
...
+		//
+		// ```Ruby
+		// 100.to_d # => '100'.to_d
+		// ```
+		// @return [Decimal]
+		{
+			Name: "to_d",
+			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
+					r := receiver.(*IntegerObject)
```
st0012: Can you add argument check here?

hachi8833: I'll do that. 

I also found that arg checks are missing in other methods so I'd add them in another PR.

vm/integer.go
```diff
@@ -383,6 +383,25 @@ func builtinIntegerInstanceMethods() []*BuiltinMethodObject {
...
+		//
+		// ```Ruby
+		// 100.to_d # => '100'.to_d
+		// ```
+		// @return [Decimal]
+		{
+			Name: "to_d",
+			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
+				return func(t *thread, args []Object, blockFrame *normalCallFrame) Object {
+					if len(args) > 1 {
```
st0012: why does it receive one argument? I don't see we use any here

hachi8833: I should've mentioned about https://github.com/goby-lang/goby/issues/648

st0012: I think either you also implement the feature that requires the argument, or you set the limit to 0 until you're going to implement that

PR #650, opened 2018-04-14
Add Class#rap?, which is equivalent to Ruby's Object#respond_to?

hachi8833:
I think the shorter method name `Class#rap?` is better Ruby's `#respond_to?`.

If most people needs `#responds_to?`, I think just aliasing to it is sufficient.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -553,6 +553,38 @@ func builtinClassCommonClassMethods() []*BuiltinMethodObject {
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


st0012:
@hachi8833 actually I think `respond_to?` is better because it's more clear about what it does. And I also think that just change the name to `respond_to?` will be good, because keeping both `rap?` and `respond_to?` violates Goby's one practice principle.

hachi8833:
OK, I will rename the method 😄

PR #658, opened 2018-04-29
Add Ripper class

hachi8833:
I'm trying to add `Ripper` loadable library like Ruby's one http://ruby-doc.org/stdlib-2.5.0/libdoc/ripper/rdoc/Ripper.html

I think `Ripper` will be needed to build Goby's linters in someday.

- Tests are needed
- Wondering how to print sexp or like that

Hope someone would help or implement those.

### DIFF
--- a/native/ripper/ripper.go
+++ b/native/ripper/ripper.go
@@ -1,0 +1,340 @@
+package ripper
+
+import (
+	"github.com/goby-lang/goby/compiler"
+	"github.com/goby-lang/goby/compiler/bytecode"
+	"github.com/goby-lang/goby/compiler/lexer"
+	"github.com/goby-lang/goby/compiler/parser"
+	"github.com/goby-lang/goby/compiler/token"
+	"github.com/goby-lang/goby/vm"
+	"github.com/goby-lang/goby/vm/classes"
+	"github.com/goby-lang/goby/vm/errors"
+	"strings"
+)
+
+// Ripper is a loadable library and has abilities to obtain parsed/lexed/tokenized Goby codes from String.
+// The library would be convenient for validating Goby codes when building lint tools,
+// as well as the tests for Goby's compiler.
+// For now, Ripper is a class and has only class methods, but I think this should finally be a 'newable' module with more sophisticated instance methods.
+
+// Imported objects from vm
+type Object = vm.Object
+type VM = vm.VM
+type Thread = vm.Thread
+type Method = vm.Method
+type StringObject = vm.StringObject
+type HashObject = vm.HashObject
+type ArrayObject = vm.ArrayObject
+
+// Class methods --------------------------------------------------------
+
+// Returns the list of instruction code generated from Goby code.
+// Returns an error when the code is invalid.
+// The return value is a "tuple" style nested array:
+// - `Array`: contains an instruction set
+//   - `arg_types:` (none if `nil`)
+// 		 - `names:` array of names (string)
+//     - `types:` array of types (integer)
+//   - `instructions:` array of instructions
+//     - `action:` string
+//     - `anchor:` integer
+//     - `line:` integer
+//     - `params:` array of parameters (string)
+//     - `source_line:` integer
+//     - `arg_set:` (none if `nil`)
+//   		 - `names:` array of names (string)
+//       - `types:` array of types (integer)
+//
+// ```ruby
+// require 'ripper'; Ripper.instruction "10.times do |i| puts i end"
+// #=>
+//
+// require 'ripper'; Ripper.instruction "10.times do |i| puts i" # the code is invalid
+// #=> TypeError: InternalError%!(EXTRA string=String, string=Invalid Goby code)
+// ```
+//
+// @param Goby code [String]
+// @return [Array]
+func instruction(receiver Object, sourceLine int, t *Thread, args []Object) Object {
+	if len(args) != 1 {
+		return t.VM().InitErrorObject(errors.ArgumentError, sourceLine, "Expect 1 argument. got=%d", len(args))
+	}
+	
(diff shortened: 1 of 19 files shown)


hachi8833:
`lex`, `parse`, `token` have been implemented.
I'd try to implement `instruction` that returns a nested `Hash`.

I hope the `Ripper` can finally be a loadable module with instance methods. How is this?

st0012:
@hachi8833 why should it be a module instead of class? can you provide a pseudo code snippet about how should we use the Ripper?

hachi8833:
As you know, Goby's modules can be instantiated with `new` like a class, while Ruby's modules cannot be done like that.

I suppose that the fact would be good to promote reducing unnecessary inheritances in Goby's codes.

I'd just like to try to build `Ripper` as a native "module" that can be `included` or instantiated as the example of that, though I haven't imagined about the real use-cases so much.

```ruby
# using `include`: this can also be done with inheritance with `< Ripper` if Ripper is a class
require `ripper`

class Foo
  include Ripper
  include SomeOtherModule
  ...

  def do_something(string)
  end
...
end
```

```ruby
# using composition: you can also inject modules to the class if needed
require 'ripper'

class Foo
  attr_reader :code
  def initialize(string)
    @code = Ripper.new(string)
  end

  def do_something
    ary = code.token
    # do something
  end
...
end
```

Of course, they are all possible if `Ripper` is a class. Only the difference is between using `<` or using `include`/`extend`.

Please let me know if I misunderstand or look over something. If you think making `Ripper` as a module is redundant, I keep it as is.

hachi8833:
Added the `instruction` method.

st0012:
@hachi8833 I think we can wait for #662 before continuing here's work. Because the Ripper class should be extracted as an external library and the structure will be different.

hachi8833:
That makes sense! I hold this on.

hachi8833:
Updated the `Ripper` class to fit #662

st0012:
@hachi8833 is it possible to move this to `/native` like https://github.com/goby-lang/goby/blob/master/native/db/db.go?

hachi8833:
I'm moving `Ripper` class under native/ripper directory but stuck on resolving `t.VM().InitStringObject` within the directory:

20180613_085405_cbj5pq

This is the same in `DB` class under native/db directory:

20180613_085554_njmyzk

Could you give me a help for this? > @st0012

hachi8833:
Applied the recent updates.
I just wonder how to write failing-tests 💦

native/ripper/ripper.go
```diff
@@ -0,0 +1,341 @@
+package ripper
+
+import (
+	"github.com/goby-lang/goby/compiler"
+	"github.com/goby-lang/goby/compiler/bytecode"
+	"github.com/goby-lang/goby/compiler/lexer"
+	"github.com/goby-lang/goby/compiler/parser"
+	gtoken "github.com/goby-lang/goby/compiler/token"
```
st0012: I think we can just use `token`

native/ripper/ripper.go
```diff
@@ -0,0 +1,341 @@
...
+		return t.VM().InitErrorObject(errors.ArgumentError, sourceLine, "Expect 1 argument. got=%d", len(args))
+	}
+		
+	arg, ok := args[0].(*StringObject)
+	if !ok {
+		return t.VM().InitErrorObject(errors.TypeError, sourceLine, errors.WrongArgumentTypeFormat, classes.StringClass, arg.Class().Name)
+	}
+		
+	l := lexer.New(arg.Value().(string))
+	el := t.VM().InitArrayObject([]Object{})
```
st0012: I think we should name this and `eli` more explicitly

native/ripper/ripper.go
```diff
@@ -0,0 +1,341 @@
...
+	eli := []Object{}
```
st0012: I think defining `eli` here is more for its type rather than the data itself? If so I think `var eli []Object` would be better

native/ripper/ripper.go
```diff
@@ -0,0 +1,341 @@
...
+	aType := []Object{}
+	for _, argtype := range argSet.Types() {
+		aType = append(aType, v.InitIntegerObject(argtype))
+	}
+	
+	h["types"] = v.InitArrayObject(aType)
+	return v.InitHashObject(h)
+}
+
+// TODO: This should finally be auto-generated from token.go
```
st0012: I think this should be provided from token pkg as well

native/ripper/ripper_test.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+	for i, tt := range tests {
+		evaluated := vm.ExecAndReturn(t, tt.input)
+		vm.VerifyExpected(t, i, evaluated, tt.expected)
+	}
+}
+
+//func TestRipperInstructionFail(t *testing.T) {
```
st0012: I think we can skip this

vm/errors/error.go
```diff
@@ -33,5 +33,6 @@ const (
 	WrongArgumentTypeFormat     = "Expect argument to be %s. got: %s"
 	CantYieldWithoutBlockFormat = "Can't yield without a block"
 	DividedByZero               = "Divided by 0"
+	InvalidGobyCode             = "Invalid Goby code"
```
st0012: If the Goby code is invalid, then we should return parser error. We shouldn't have this kind of error inside VM

native/ripper/ripper.go
hachi8833: I'll improve this in another PR

hachi8833:
Applied the review and finally added the failing tests.
If it is OK, I'd squash the commits before @st0012 merges this

hachi8833:
Thank you for approving this! Merging...

PR #662, opened 2018-05-05
WIP Move "db" package out of "vm"

eliothedeman:
This is a very large diff, and most of the changes are exporting internal functions so other packages can use them. In my opinion, the surface area of the API should be smaller, but for now I think exporting these functions is fine. There are also some general code quality changes as well. 

Main changes:
* Functions for registering native libraries.
* Testing helpers for external libraries.

Perf related changes include:
* VM's main thread is no longer a pointer to a thread.
* Threads no longer need any initialization logic. `var t Thread` only needs it's vm to be set before it is functional
* Thread's stack and call stack are no longer pointers but live inside of the thread itself.

```bash
➜  goby git:(thread-rework) ./benchmark.rb
benchmarking b6a1996c7832449e3c1bb71c61d6b867c91901cc
benchmarking da753973fd6100c9fc4317a9adb5c42da0d8d3d9
universal.x86_64-darwin17
benchmark                         old ns/op     new ns/op     delta
BenchmarkBasicMath/add-8          5959          5424          -8.98%
BenchmarkBasicMath/subtract-8     5918          5379          -9.11%
BenchmarkBasicMath/multiply-8     6075          5521          -9.12%
BenchmarkBasicMath/divide-8       5829          5448          -6.54%

benchmark                         old allocs     new allocs     delta
BenchmarkBasicMath/add-8          78             78             +0.00%
BenchmarkBasicMath/subtract-8     78             78             +0.00%
BenchmarkBasicMath/multiply-8     78             78             +0.00%
BenchmarkBasicMath/divide-8       78             78             +0.00%

benchmark                         old bytes     new bytes     delta
BenchmarkBasicMath/add-8          3792          3792          +0.00%
BenchmarkBasicMath/subtract-8     3792          3792          +0.00%
BenchmarkBasicMath/multiply-8     3792          3792          +0.00%
BenchmarkBasicMath/divide-8       3792          3792          +0.00%
Switched to branch 'thread-rework'
```


vm/class.go
```diff
@@ -1144,24 +1180,35 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
...
+						return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, "Expect 1 argument. got: %d", len(args))
 					}
 					switch args[0].(type) {
 					case *StringObject:
 						libName := args[0].(*StringObject).value
 						initFunc, ok := standardLibraries[libName]
 
 						if !ok {
-							return t.vm.initErrorObject(errors.InternalError, sourceLine, "Can't require \"%s\"", libName)
+							externalClassLock.Lock()
```
st0012: is this for preventing concurrent requiring?

lib/db.go
```diff
@@ -0,0 +1,327 @@
...
+import (
+	"fmt"
+
+	"github.com/goby-lang/goby/vm"
+	"github.com/goby-lang/goby/vm/errors"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
+)
+
+type Object = vm.Object
```
st0012: why do we still need to export these types from db?

lib/db.go
```diff
@@ -0,0 +1,327 @@
+package lib
```
st0012: Do you think we should create another directory to separate Go and Goby libraries?

st0012:
@eliothedeman does the performance gain come from replacing pointer with object itself?

vm/class.go
eliothedeman: Yes. Requiring happens so infrequently, I think havinga  lock is fine for now.

lib/db.go
eliothedeman: Yeah that is a good idea. What do you think about having a separate dir for each go package?

vm/method.go
```diff
@@ -75,7 +76,26 @@ type BuiltinMethodObject struct {
 	Fn   func(receiver Object, sourceLine int) builtinMethodBody
 }
 
-type builtinMethodBody func(*thread, []Object, *normalCallFrame) Object
+// Method is a callable function
+type Method = func(t *Thread, args []Object) Object
+
+// MethodBuilder constructs and instance of a method
```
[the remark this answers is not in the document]
eliothedeman: You're right.

lib/db.go
st0012: I think something like
```
lib
  - db
  - other_lib
```
would be great

lib/db.go
st0012: @eliothedeman I still don't get the reason we export these types. is there any chance for us to use `lib.Object` or `lib.VM` instead of `vm.Object` or `vm.VM` ?

eliothedeman:
@st0012 Thoughts on merging this?

st0012:
Sure if you finished it

PR #664, opened 2018-05-10
Float APIs - Part1

### LINKED ISSUE #484: VM Float API Implementation

st0012:
This PR is for continuing the work at #615 (thanks @SD10 for previous works)

It implements:
- [x] `Float#abs` Return positive number of the float. "Formally as the distance between 0 to the number of the float in 1-dimensional axis ..."
- [x] `Float#ceil` Returns the smallest number greater than or equal to the receiver
- [x] `Float#floor` Returns the largest number smaller than or equal to the receiver
- [x] `Float#round` Rounds float to a given precision in decimal digits (default 0 digits). Precision may be negative. Returns a floating-point number when `n` digits are more than zero.

- [x] `Float#positive?` Returns `true` if float number is larger than `0.0`
- [x] `Float#negative?` Returns `true` if float number is less than `0.0`
- [x] `Float#zero?` Returns `true` if float number is equal to `0.0`

### DIFF
--- a/vm/float.go
+++ b/vm/float.go
@@ -320,6 +320,70 @@ var builtinFloatInstanceMethods = []*BuiltinMethodObject{
 
 		},
 	},
+	{
+		// Returns the Float as a positive value.
+		//
+		// ```Ruby
+		// -3.14.abs # => 3.14
+		// 3.14.abs # => 3.14
+		// ```
+		// @return [Float]
+		Name: "abs",
+		Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
+			if len(args) != 0 {
+				return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, "Expect 0 argument. got=%v", strconv.Itoa(len(args)))
+			}
+			r := receiver.(*FloatObject)
+			result := math.Abs(r.value)
+			return t.vm.initFloatObject(result)
+		},
+	},
+	{
+		// Returns the smallest Integer greater than or equal to self.
+		//
+		// ```Ruby
+		// 1.2.ceil # => 2
+		// 2.ceil # => 2
+		// -1.2.ceil # => -1
+		// -2.ceil # => -2
+		// ```
+		// @return [Integer]
+		Name: "ceil",
+		Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
+			// TODO: Make ceil accept arguments
+			if len(args) != 0 {
+				return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, "Expect 0 argument. got=%v", strconv.Itoa(len(args)))
+			}
+			r := receiver.(*FloatObject)
+			result := math.Ceil(r.value)
+			newInt := t.vm.InitIntegerObject(int(result))
+			newInt.flag = i
+			return newInt
+		},
+	},
+	{
+		// Returns the largest Integer less than or equal to self.
+		//
+		// ```Ruby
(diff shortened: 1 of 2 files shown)


vm/float.go
```diff
@@ -317,7 +317,239 @@ var builtinFloatInstanceMethods = []*BuiltinMethodObject{
...
+		// Returns the largest Integer less than or equal to self.
+		//
+		// ```Ruby
+		// 1.2.floor  # => 1
+		// 2.0.floor  # => 2
+		// -1.2.floor # => -2
+		// -2.0.floor # => -2
+		// ```
+		// @return [Integer]
+		Name: "floor",
```
saveriomiroddi, line 437: It seems this is duplicated?

vm/float.go
```diff
@@ -317,7 +317,239 @@ var builtinFloatInstanceMethods = []*BuiltinMethodObject{
...
+		// Returns the smallest Integer greater than or equal to self.
+		//
+		// ```Ruby
+		// 1.2.ceil  # => 2
+		// 2.ceil    # => 2
+		// -1.2.ceil # => -1
+		// -2.ceil   # => -2
+		// ```
+		// @return [Integer]
+		Name: "ceil",
```
saveriomiroddi, line 414: I think this is duplicated, too.

vm/float.go
```diff
@@ -317,7 +317,239 @@ var builtinFloatInstanceMethods = []*BuiltinMethodObject{
...
+	},
+	{
+		// Returns the Float as a positive value.
+		//
+		// ```Ruby
+		// -34.56.abs # => 34.56
+		// 34.56.abs # => 34.56
+		// ```
+		// @return [Float]
+		Name: "abs",
```
saveriomiroddi, line 394: Third duplicate :grimacing:

vm/float.go
st0012: rebasing 🤦

SD10:
🎉 Stay safe and healthy my friends

PR #666, opened 2018-05-13
[WIP]Feature/issue 660

### LINKED ISSUE #660: add numbers with underscore
ruby had this nice feature to add underscore between digits for numbers `1_000_000_000`
it would be great if goby can do it too.

shes50103:
- Implement underscore feature
- make sure underscore using in the right way

`1_2_3` => 123
`1_2.3_4` => 12.34
`1__2` => Error
`_1` => Error
`1_` => Error 

This closes #660

### DIFF
--- a/compiler/lexer/lexer.go
+++ b/compiler/lexer/lexer.go
@@ -264,6 +264,13 @@ func (l *Lexer) resetNosymbol() {
 
 }
 
+// prevent double underscores and underscore at the end such as 1__2, 3_
+func (l *Lexer) checkNumberError() bool {
+
+	return l.ch == '_' && (l.readPosition >= len(l.input) || !isDigit(l.input[l.readPosition]))
+
+}
+
 func (l *Lexer) readNumber() []rune {
 	var input []rune
 
@@ -271,6 +278,10 @@ func (l *Lexer) readNumber() []rune {
 
 		if isDigit(l.ch) {
 			input = append(input, l.ch)
+		}
+
+		if l.checkNumberError() {
+			return input
 		}
 
 		l.readChar()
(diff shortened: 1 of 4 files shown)


compiler/lexer/lexer.go
```diff
@@ -264,12 +264,29 @@ func (l *Lexer) resetNosymbol() {
...
+}
+
 func (l *Lexer) readNumber() []rune {
-	position := l.position
-	for isDigit(l.ch) {
+	var input []rune
+
+	for isDigit(l.ch) || l.ch == '_' {
+
+		if isDigit(l.ch) {
```
SD10: IMO I think the lexer should preserve the underscores in the token. In terms of responsibility, I don't think a lexer should be responsible for any code transformation. Removing them at this stage seems sort of magical. 

IMO the token should be simplified at a later stage, maybe during parsing or eval?

st0012:
Well I think @SD10 has a good point. Keeping those information to parser allows us to provide more sophisticated error message

SD10:
Here we are already transforming the token into a literal for `Int`:
https://github.com/goby-lang/goby/blob/c00ad0ca3d6b779b6ab737dea906637ac62e4708/compiler/parser/data_type_parsing.go#L12
Here for Float:
https://github.com/goby-lang/goby/blob/c00ad0ca3d6b779b6ab737dea906637ac62e4708/compiler/parser/data_type_parsing.go#L26

This is just my opinion. I'm a newbie 😊

SD10:
It's also worth noting that in the Ruby lexer the underscores are maintained:
```
...
[[2, 7], :on_int, "1_000_000"],
```

PR #669, opened 2018-05-15
Update api-doc in boolean.go

hachi8833:
I also found that the instance methods in `Boolean` class are completely redundant.
Removed them and updated `Class#!`.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -676,11 +676,13 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 			},
 		},
 		{
-			// Inverts the boolean value.
+			// Inverts the boolean value. Any objects other than `nil` and `false` are `true`, thus returns `false`.
 			//
 			// ```ruby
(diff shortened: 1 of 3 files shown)


vm/class.go
```diff
@@ -676,11 +676,13 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
 			},
 		},
 		{
-			// Inverts the boolean value.
+			// Inverts the boolean value. Any objects other than `nil` and `false` are `true`, thus returns `false`.
 			//
 			// ```ruby
 			// !true  # => false
 			// !false # => true
+			// !nil   # => true
```
st0012: Do we have test case for this condition? According to the code below it looks like this will return false because it's not a boolean object

hachi8833: Added evaluation tests for that.
`!nil` works as expected to be `true`: the negation of `false` that is equivalent to`nil`

PR #670, opened 2018-05-15
Update api-doc for channel.go

hachi8833:
I hope my understanding is correct, especially the reason why argument checking is omitted.

### DIFF
--- a/vm/channel.go
+++ b/vm/channel.go
@@ -2,12 +2,61 @@ package vm
 
 import (
 	"fmt"
-	"sync"
-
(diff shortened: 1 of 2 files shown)


st0012:
@hachi8833 well actually we don't have any particular reason to avoiding argument check on `#receive`, just didn't put the checking logic 😅

hachi8833:
Thank you for checking this!
I'd update them to add argument checking after merging this

st0012:
@hachi8833 I think you can take this as a reference: https://tour.golang.org/concurrency/4

PR #671, opened 2018-05-19
Add argument/status checking to methods in Channel class

hachi8833:
I found that the methods in `Channel` were unsafe:

- when `close` is called twice
- when `receive` is called against the closed channel
- when `deliver` is called against the closed channel

Using a status flag `ChannelState` might look unsophisticated. I'd be glad if someone else have a better idea for this.
But I think the methods are now safer than before, or it might be safer than raw Golang 😆 

Well, I'm wondering how to release the ChannelObject when `closed`.


vm/channel.go
```diff
@@ -63,9 +63,22 @@ import (
...
+// Channel's state.
+// To Goby language contributors: Golang's channels should be carefully handled because:
+// - You cannot write to closed channels, or got a panic.
+// - You cannot close closed channels, or got a panic.
+// - You cannot write to nil channels, or causes a deadlock.
+// - You cannot read nil channels, or causes a deadlock.
+// Ref: https://beatsync.net/main/log20150325.html
+const (
+	OPEN   int = iota
+	CLOSED int = iota
```
st0012: This `iota` is redundant I think

vm/channel.go
```diff
@@ -118,16 +130,26 @@ func builtinChannelInstanceMethods() []*BuiltinMethodObject {
...
 			Fn: func(receiver Object, sourceLine int) builtinMethodBody {
 				return func(t *Thread, args []Object, blockFrame *normalCallFrame) Object {
+					if len(args) != 0 {
+						return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, "Expect 0 argument. got: %d", len(args))
+					}
 					c := receiver.(*ChannelObject)
 
-					close(c.Chan)
+					if c.ChannelState == CLOSED {
+						return t.vm.InitErrorObject(errors.InternalError, sourceLine, "The channel is already closed. got: %d", len(args))
```
st0012: I think we should have a dedicated error type for this

hachi8833: Got it

vm/channel.go
hachi8833: I'd remove the second one

vm/channel.go
hachi8833: Updated

vm/channel.go
hachi8833: Updated

vm/channel.go
```diff
@@ -63,9 +63,22 @@ import (
...
 
+// Channel's state.
+// To Goby language contributors: Golang's channels should be carefully handled because:
+// - You cannot write to closed channels, or got a panic.
+// - You cannot close closed channels, or got a panic.
+// - You cannot write to nil channels, or causes a deadlock.
+// - You cannot read nil channels, or causes a deadlock.
+// Ref: https://beatsync.net/main/log20150325.html
+const (
+	OPEN = iota
```
eliothedeman: Probably want to prefix with some identifier that these refer to channel states. Also, prefer the go convention of using camel case. These also probably shouldn't be exported values. Packages outside of vm should not be touching these state values.

hachi8833: Thank you! I missed the point.

vm/channel.go
```diff
@@ -13,7 +14,6 @@ import (
 // `Channel#new` is available.
 //
 // Note that channels are not like files and you don't need to explicitly close them (e.g.: exiting a loop).
-// If you `close` the closed channel, Goland will cause a panic. TODO: this panic should be avoided.
```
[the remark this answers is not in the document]
hachi8833: I'll fix this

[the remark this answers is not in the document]
hachi8833: Updated the sentence

vm/channel.go
hachi8833: Sorry for the delay: updated the constants

hachi8833:
I think it's ready to merge. Could you do that?

PR #672, opened 2018-05-21
Support method missing

st0012:
Not support `method_missing`'s inheritance but support singleton class' `method_missing`

### DIFF
--- a/vm/instruction.go
+++ b/vm/instruction.go
@@ -207,7 +207,7 @@ var builtinActions = map[operationType]*action{
 		name: bytecode.NewArray,
 		operation: func(t *Thread, sourceLine int, cf *normalCallFrame, args ...interface{}) {
 			argCount := args[0].(int)
-			elems := []Object{}
+			var elems []Object
 
 			for i := 0; i < argCount; i++ {
 				v := t.Stack.Pop()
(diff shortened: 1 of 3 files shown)


vm/instruction.go
```diff
@@ -496,7 +496,28 @@ var builtinActions = map[operationType]*action{
...
+				if mm == nil {
+					t.setErrorObject(receiverPr, argPr, errors.UndefinedMethodError, sourceLine, "Undefined Method '%+v' for %+v", methodName, receiver.toString())
+				} else {
+					// Move up args for missed method's name
+					// before: | arg 1       | arg 2 |
+					// after:  | method name | arg 1 | arg 2 |
+					// TODO: Improve this
+					t.Stack.Push(&Pointer{Target: nil})
+
+					for i := argCount - 1; i >= 0; i-- {
```
st0012: @eliothedeman @saveriomiroddi  Because we need to append method's name as first argument, so we have an O(n) operation here. I'd be very happy if anyone can solve this elegantly

hachi8833:
I just wonder if `__method_missing__` or `method_missing__` might be good for avoiding exploitation. How about this?

vm/instruction.go
```diff
@@ -604,39 +625,39 @@ var builtinActions = map[operationType]*action{
 	},
 }
 
-func (vm *VM) InitObjectFromGoType(value interface{}) Object {
-	switch v := value.(type) {
+func (v *VM) InitObjectFromGoType(value interface{}) Object {
```
eliothedeman: Thank you!

PR #673, opened 2018-05-23
Refactor error message constants (starting)

hachi8833:
Excuse me for the wide update.

While updating API documents, I found that several inconsistent and literal messages have been found across the codes.

This is the first step to integrate them. I'd continue to do until all the error messages are consistent.

If the refactoring is OK, I'd squash and push them again


vm/errors/error.go
```diff
@@ -21,14 +21,45 @@ const (
 	HTTPError = "HTTPError"
 	// ZeroDivisionError is for zero-division by Integer/Float/Decimal value
 	ZeroDivisionError = "ZeroDivisionError"
+	// BlockError is for indicating errors regarding block
+	BlockError = "BlockError"
+	// FileError is for indicating errors regarding files
+	FileError = "FileError"
```
st0012: I think these kind of error should be contained in the class file

vm/errors/error.go
```diff
@@ -21,14 +21,45 @@ const (
...
-*/
+// Error messages for native methods
 const (
-	WrongNumberOfArgumentFormat = "Expect %d arguments. got: %d"
-	WrongArgumentTypeFormat     = "Expect argument to be %s. got: %s"
+	WrongNumberOfArgumentFormat      = "Expects %d argument(s). got: %d"
+	WrongNumberOfArgumentFormatMore  = "Expects %d or more argument(s). got: %d"
+	WrongNumberOfArgumentFormatRange = "Expects %d to %d argument(s). got: %d"
+	WrongArgumentTypeFormat          = "Expects argument to be %s. got: %s"
+	WrongArgumentSignNegative        = "Expects argument to be positive. got: %d"
```
st0012: Do we need an error this specific?

vm/errors/error.go
```diff
@@ -21,14 +21,45 @@ const (
...
+
+	IndexValueTooSmall = "Index value %d is too small for array. minimum: %d"
```
st0012: I think this error should be put inside `array.go`

hachi8833:
> Some type/library specific errors or messages shouldn't be contain in the more general `errors` package I think

Reorganized the errors and messages so that class-specific errors are contained within the class

For now, only array.go has been updated. the rest are in progress.
I'd be glad if you are OK with this and thenmerge first then I continue updating the rest of the classes under vm > @st0012

vm/array.go
```diff
@@ -18,6 +18,14 @@ type ArrayObject struct {
 	splat    bool
 }
 
+const (
+	// ArrayError is for array-specific error
+	ArrayError = "ArrayError"
```
st0012: I think this should be unexpected

st0012: Otherwise it's good to go

hachi8833: I'd check around this

PR #681, opened 2018-06-09
Prohibit module initialization

### LINKED ISSUE #679: Undocumented: Goby's modules can be `new`
Actually, I found the undocumented behavior some months ago, and the st0012 and I left this as is by now.

Goby's modules can be instantiated:

```ruby
#Goby
module Foo
end

Foo.new       # modules can be instantiated in Goby
#» <Instance of: Foo>

# of course inheritance of modules is prohibited
class Bar < Foo
end
#» InternalError: Module inheritance is not supported: Foo
```

Of course this is prohibited in Ruby.
But I think the behavior suggests using modules more frequent rather than inheritance so I prefer to it.

Just close the issue if this is OK > @st0012 
If not, let get started a discussion.

st0012:
This fixes #679. The solution is more complicated than just forbidden the method. I choose to let it work more like Ruby:

1. `Class < Module` and `Module < Object`
2. Module class doesn't support `#new` (but support `.new`)
3. Class class support `#new`

Before today our inheritance chain for `Class` class is `Class < Object`, so I changed it as described above.

@hachi8833 can you help me compare the Goby's behavior with Ruby again?

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -1399,7 +1399,7 @@ func (vm *VM) InitArrayObject(elements []Object) *ArrayObject {
 }
 
 func (vm *VM) initArrayClass() *RClass {
-	ac := vm.initializeClass(classes.ArrayClass, false)
+	ac := vm.initializeClass(classes.ArrayClass)
 	ac.setBuiltinMethods(builtinArrayInstanceMethods(), false)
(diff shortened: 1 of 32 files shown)


st0012:
@hachi8833 can you review and approve it if looks ok?

PR #684, opened 2018-06-13
Export InitStringObject function

st0012:


st0012:
@hachi8833 We can wait till we need them

PR #688, opened 2018-06-19
Use dep instead of godep

eliothedeman:
As stated in the readme https://github.com/tools/godep Godeps is no longer supported. This migrates to use dep, which is the officially supported dependency management tool


st0012:
@eliothedeman can you delete `Godep` directory as well?

PR #696, opened 2018-07-03
NullObject now converts to a blank string in all the cases

### LINKED ISSUE #695: Printing nil should not print anything
Currently, printing (via `puts()`) nil will print `nil`. This is not inline with Ruby, and especially, not convenient, since it breaks patterns like:

```ruby
infix = "infix" if conditional
string = "prefix#{infix}suffix"
```

saveriomiroddi:
Previously, the internal NullObject#toString would convert to `nil`, which is not correct.

Closes #695.

Will need some rework when #698 is implemented.

### DIFF
--- a/vm/array_test.go
+++ b/vm/array_test.go
@@ -144,7 +144,7 @@ func TestArrayIndex(t *testing.T) {
 			code[102] = 'Processing'
 			code[200] = 'OK'
 			code.to_s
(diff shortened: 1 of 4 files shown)


st0012:
@saveriomiroddi this breaks one of the Array's test (which I think needs some rewrite)

PR #700, opened 2018-07-07
Move plugin out of vm package

eliothedeman:


native/plugin/plugin_integration_test.go
```diff
@@ -31,15 +30,12 @@ func TestCallingPluginFunctionWithReturnValueNoRaceDetection(t *testing.T) {
 	input := `
 	require "plugin"
 
-	p = Plugin.use "../test_fixtures/import_test/plugin/plugin.go"
+	p = Plugin.use "../test_fixtures/import_test.VM()/plugin.go"
```
st0012: Why changing the path, did you move the file as well?

eliothedeman: bad search and replace

hachi8833:
Thank you! This might help finish https://github.com/goby-lang/goby/pull/658 under way.

PR #701, opened 2018-07-07
Binder

eliothedeman:
This is my first pass at writing a binding generator for goby. This allows you to write goby code as regular go structs, and takes care of the annoying bit of mapping that to goby methods. I'm still debugging an internal panic that is being thrown, but the code functionally works.


native/result/bindings.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	arg0, ok := args[0].(Object)
+	if !ok {
+		panic("Argument 0 must be Object")
+	}
+
+	return r.MethodMissing(t, arg0)
+}
+
+func _binding_Result_Or(receiver vm.Object, line int, t *vm.Thread, args []vm.Object) vm.Object {
```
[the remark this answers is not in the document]
st0012: @eliothedeman will we change the function name to use more Go-like naming?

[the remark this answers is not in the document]
eliothedeman: This is tough, because I don't want to have to build logic for handling name conflicts. So using a name like this makes it pretty likely that won't happy.

native/result/result_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func TestEmptyResult(t *testing.T) {
+	r := vm.ExecAndReturn(t, `
+		require("result")
+		r = Result.empty()
+	`)
+
+	if r, ok := r.(*Result); ok {
+		if !r.empty {
```
st0012: I don't think we should do verification in Go, we can add something like `#_empty?` as test helper. I actually prefer writing these cases using builtin test framework

native/result/bindings.go
```diff
@@ -0,0 +1,82 @@
...
+
+// bindingResultNew is genderated by github.com/goby-lang/goby/cmd/binder
+func bindingResultNew(receiver vm.Object, line int, t *vm.Thread, args []vm.Object) vm.Object {
+	r := staticResult
+	if len(args) != 2 {
+		return t.VM().InitErrorObject(errors.ArgumentError, line, errors.WrongNumberOfArgumentFormat, 2, len(args))
+	}
+	arg0, ok := args[0].(Object)
+	if !ok {
+		panic("Argument 0 must be Object")
```
st0012: is this check necessary?

native/result/bindings.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	arg1, ok := args[1].(Object)
+	if !ok {
+		panic("Argument 1 must be Object")
```
st0012: and this one

cmd/binder/main.go
```diff
@@ -0,0 +1,290 @@
...
+			Line().Map(String()).Qual(vmPkg, "Method").Values(im),
+		),
+	)
+	l := Func().Id("init").Params().Block(
+		dm,
+	)
+	return l
+}
+
+func main() {
```
st0012: Can you add binder's usage document above?

native/result/result_test.go
st0012: Maybe we can use a Go test function to start Goby spec then their coverage will be included?

cmd/binder/main.go
st0012: I mean like what's the scenario to use it, what do you need for using it and what options do you have..etc.

st0012:
@eliothedeman is this finished?

eliothedeman:
@st0012 Yeah this is ready to merge whenever you're ready.

PR #704, opened 2018-07-22
Implement Array#sort

st0012:
I use Golang's `sort` package for the sorting, so the change is for exposing the required interface


hachi8833:
I have a question: how do you want to organize the methods for sorting in the future?

* Current `Array#sort` is very simple and takes no arguments including a block. 
* Currently we don't have `<=>` operator and `Enumerable` module and `Enumerator` class like Ruby.
* Instead we have `ArrayEnumerator`, `LazyEnumerator`, `RangeEnumerator` classes as well as `Hash#sort` and `Hash#sorted_key` methods

Regards,

hachi8833:
I understood! I expect the discussions for the better design around this 😋

PR #705, opened 2018-07-22
Use a simpler concurrency benchmark script.

st0012:
@eliothedeman can you help me on this? I'm not sure what script is best for benchmarking concurrency


st0012:
@eliothedeman I'll say relatively consistent because they're so slow and allocate a lot (`2731306` bytes for example). That means the number will go up and down during every run

PR #708, opened 2018-07-24
Remove unnecessary conversions on instructions

st0012:
We used to convert all types of data into string when compiling instructions, and convert them back to their original data type in vm. This was for the goal of compiling to binary, and since we're not going to do that anymore we should skip this conversion.

And as @eliothedeman said before, this does improve the performance (100% improvement!):

```
benchmarking 0772826a0a295544b55e35b10f10877af452355e
benchmarking a200a8af3b0cf0a7f4546291a17e45c4a8b99078
x86_64-darwin17
benchmark                         old ns/op     new ns/op     delta
BenchmarkBasicMath/add-4          5139          2810          -45.32%
BenchmarkBasicMath/subtract-4     5206          2798          -46.25%
BenchmarkBasicMath/multiply-4     5158          2813          -45.46%
BenchmarkBasicMath/divide-4       5184          2807          -45.85%

benchmark                         old allocs     new allocs     delta
BenchmarkBasicMath/add-4          75             34             -54.67%
BenchmarkBasicMath/subtract-4     75             34             -54.67%
BenchmarkBasicMath/multiply-4     75             34             -54.67%
BenchmarkBasicMath/divide-4       75             34             -54.67%

benchmark                         old bytes     new bytes     delta
BenchmarkBasicMath/add-4          3152          1576          -50.00%
BenchmarkBasicMath/subtract-4     3152          1576          -50.00%
BenchmarkBasicMath/multiply-4     3152          1576          -50.00%
BenchmarkBasicMath/divide-4       3152          1576          -50.00%
```


native/ripper/ripper_test.go
```diff
@@ -300,76 +300,76 @@ func TestRipperLexFail(t *testing.T) {
...
-	}
-	for i, tt := range tests {
-		evaluated := vm.ExecAndReturn(t, tt.input)
-		vm.VerifyExpected(t, i, evaluated, tt.expected)
-	}
-}
+//func TestRipperInstruction(t *testing.T) {
```
st0012: @hachi8833 can you take a look at this when you got a chance? It's too verbose to fix so I have to comment them out for now

hachi8833: Got it

hachi8833: Looks the recent commits might affect this. I will check this again when your refactoring finishes.

st0012: Ok I finished changes, you can check that when you're free

compiler/bytecode/instruction.go
```diff
@@ -16,55 +15,104 @@ const (
...
+	InvokeBlock
+	GetBlock
+	Pop
+	Dup
+	Leave
 )
 
+
+// InstructionNameTable is the table the maps instruction's op code with its readable name
+var InstructionNameTable = map[uint8]string{
```
eliothedeman: Prefer using [...]string{} for things like this as you are just mapping integers 0...n to strings.

vm/instruction_translator.go
```diff
@@ -44,87 +41,18 @@ func (it *instructionTranslator) setMetadata(is *instructionSet, set *bytecode.I
...
-		return param
-	}
-
-	i := int(integer)
-
-	return i
-}
-
 func (it *instructionTranslator) transferInstructionSets(sets []*bytecode.InstructionSet) []*instructionSet {
 	iss := []*instructionSet{}
```
eliothedeman: Can save a few allocations here by doing make([]*instructionSeet, 0, len(sets)). Also, do the instructionSets need to be pointers? Mind you, that each one will be a separate allocation.

compiler/bytecode/instruction.go
st0012: 👍

vm/instruction_translator.go
st0012: Anyway, this is actually redundant so removing it reduced 1 additional allocation 😄

PR #709, opened 2018-07-29
Fix #702 -- Update error messages for NoMethodError

### LINKED ISSUE #702: Integer::new has weird behavior
I tried calling Integer.new just for an experiment, and I got `UnsupportedMethodError: Unsupported Method #new for Integer` as an error.. I think that is misleading. What is an "unsupported method", and why should unsupported methods exist at all? Isn't it better to remove it from the method list of the Integer class (and thus also not display it in ::methods)?

In Ruby, this simply raises ```NoMethodError: undefined method `new' for Integer:Class```

hachi8833:
Fixes #702.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -99,7 +99,7 @@ func builtinClassCommonClassMethods() []*BuiltinMethodObject {
 				class, ok := receiver.(*RClass)
 
 				if !ok {
-					return t.vm.InitUnsupportedMethodError(sourceLine, "#new", receiver)
+					return t.vm.InitNoMethodError(sourceLine, "#new", receiver)
 				}
 
 				instance := class.initializeInstance()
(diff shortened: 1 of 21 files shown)


vm/class.go
```diff
@@ -99,7 +99,7 @@ func builtinClassCommonClassMethods() []*BuiltinMethodObject {
 				class, ok := receiver.(*RClass)
 
 				if !ok {
-					return t.vm.InitUnsupportedMethodError(sourceLine, "#new", receiver)
+					return t.vm.InitNoMethodError(sourceLine, "new", receiver)
```
st0012: I think we can still use `InitNoMethodError` here (and below)?

hachi8833: I'd check around this

hachi8833: Applied line 159 and some more.

PR #714, opened 2018-08-08
Add CGO_ENABLED=0 to supress dynamic linking

hachi8833:
I found that the Goby's binary executable  has been partially dynamic-linked.
Adding `CGO_ENABLED=0` to Makefile to ensuring static-linking.

Ref http://blog.wrouesnel.com/articles/Totally%20static%20Go%20builds/


st0012:
@hachi8833 But for using plugin we need cgo enabled [reference](https://medium.com/@diogok/on-golang-static-binaries-cross-compiling-and-plugins-1aed33499671#a4bf)

hachi8833:
I missed the point 😓 
Closing the PR...

PR #718, opened 2018-08-17
bugfix: allow more than 5 local variables

mweitzel:
Bugfix: allow more than 5 local variables.

The method `insertLCL` auto-expands appropriately, but calls to `getLCL` attempt to read outside the bounds of the current array.

introduced in 8c3addf0f40a4bc71484c9461670e3602419c625

### DIFF
--- a/vm/call_frame.go
+++ b/vm/call_frame.go
@@ -128,7 +128,9 @@ func (b *baseFrame) getLCL(index, depth int) (p *Pointer) {
 
 	if depth == 0 {
 		b.RLock()
-		p = b.locals[index]
+		if index < len(b.locals) {
+			p = b.locals[index]
+		}
 		b.RUnlock()
 		return
 	}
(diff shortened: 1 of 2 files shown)


mweitzel:
@st0012 I was going to do that, but it would be checking outside the lock. Should I add it there and move the lock?

vm/call_frame.go
```diff
@@ -128,7 +128,9 @@ func (b *baseFrame) getLCL(index, depth int) (p *Pointer) {
 
 	if depth == 0 {
 		b.RLock()
-		p = b.locals[index]
+		if index < len(b.locals) {
```
st0012: Or I think we can change this to
```go
if index >= len(b.locals) {
  return
}
p = b.locals[index]
```

so we can explicitly tell others that we intend to return nil in such situation. The original way looks like `p = b.locals[index]` is a special condition because it's put inside an if statement

mweitzel: Cool, thanks. Will do.

vm/fmt/fmt.go
```diff
@@ -0,0 +1 @@
+package fmt
```
st0012: @mweitzel what's this?

st0012:
@mweitzel great job, thanks!

PR #722, opened 2018-08-22
Improve case statement and REPL

hachi8833:
I realized that Ruby's `case` permits empty conditions:

```ruby
[1] pry(main)> case
[1] pry(main)* when 1
[1] pry(main)*   1
[1] pry(main)* when 2
[1] pry(main)*   2
[1] pry(main)* end
=> 1
```

Thus I have made the change to be like above:

```ruby
Goby 0.1.10 😌 👺 😻
» class Foo
¤   def bar
¤     case
¤     when 1
¤       99
¤     when 2
¤       if true
¤         88
¤       elsif false
¤         77
¤       else
¤         66
¤       end
¤     end
¤     case
¤     when 7
¤       55
¤     when 8
¤       44
¤     when 9
¤       33
¤     end
¤   end
» end
#»
```

Actually, previous REPL had an issue that it cannot handle `case` sentences well. I also fixed this and now works fine as above.

### DIFF
--- a/compiler/parser/errors/errors.go
+++ b/compiler/parser/errors/errors.go
@@ -57,6 +57,11 @@ func (e *Error) IsUnexpectedWhen() bool {
 	return e.ErrType == UnexpectedTokenError && len(e.Message) >= 21 && e.Message[0:21] == "unexpected when Line:"
 }
 
+// IsUnexpectedEmptyLine checks if error is unexpected 'end' with empty line
+func (e *Error) IsUnexpectedEmptyLine(len int) bool {
+	return e.IsUnexpectedEnd() && len == 0
+}
+
 // InitError is a helper function for easily initializing error object
 func InitError(msg string, errType int) *Error {
 	return &Error{Message: msg, ErrType: errType}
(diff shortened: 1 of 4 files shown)


compiler/parser/errors/errors.go
```diff
@@ -41,6 +42,21 @@ func (e *Error) IsUnexpectedEnd() bool {
...
 }
 
+// IsUnexpectedTok checks if error is unexpected token error
+func (e *Error) IsUnexpectedToken() bool {
+	return e.ErrType == UnexpectedTokenError
+}
+
+// IsUnexpectedCase checks if error is unexpected token error for 'case'atement
+func (e *Error) IsUnexpectedCase() bool {
+	return e.ErrType == UnexpectedTokenError && len(e.Message) >= 49 && e.Message[0:49] == "expected next token to be WHEN, got EOF() instead"
```
st0012: I think what we should do is to include the wrong token in the `Error`. since most of the parser error exists in the expected token form

hachi8833: Makes sense!
I found some more minor issues in the PR and will update soon.

hachi8833: fixed

hachi8833:
I also found that the PR fixes other potential panic on REPL.
The following works on the PR.

```ruby
block = Block.new do
  def new_method
    self.reverse
  end
end
str = "String"
a = str.instance_eval(block)  #=> panic on current master
```

I think this PR is ready to merge.

PR #725, opened 2018-08-25
Use stringer to reduce duplicate constants in bytecode instruction.go

hachi8833:
Finally I found the way to use `stringer` to remove duplicate constants.
Updated the Make file to add `make generate` (I wonder there might be a better way)

If this is OK, I'd try using `stringer` against token.go.


st0012:
I think this is a great attempt but not seeing as much benefit as the complexity it adds. @eliothedeman what do you think about this?

eliothedeman:
Code generation is pretty standard practice for interpreters, so I'm not generally opposed. I'll have to think about it a bit more though.

st0012:
@hachi8833 thanks for the effort! But I think at this stage, maintainability is more important than optimization. And since it's very likely that we'll still add/remove instructions, I think we shouldn't do it now. So I'm closing it now, can revisit in the future.

PR #729, opened 2018-08-26
Fix #726: float `0.0` should not be truncated to `0`

### LINKED ISSUE #726: Bug: ".0000" in Float and Decimal are truncated
`Float` literals that ends with `.0` (or `.00` or more) are all dropped while they are still `Float`:

```ruby
a = [1, 2, 3, 4.1, 9.000]
a  #=> [1, 2, 3, 4.1, 9]       -- the last element should be 9.0
a[4]           #=> 9           -- this should be 9.0
a[4].class  #=> Float
```

Same as the case of `Decimal`:

```ruby
a = [1, 2, 3, "4.1".to_d, "9.000".to_d]
a  #=> [1, 2, 3, 4.1, 9]       -- the last element should be 9.0
a[4]        #=> 9              -- this should be 9.0
a[4].class  #=> Decimal
```

To avoid confusing developers, `.0` should always be preserved for `Float` and `Decimal`.

I should've noticed when implementing `Decimal` class.
I will looking into it.

hachi8833:
Fixes https://github.com/goby-lang/goby/issues/726

(The behavior of Decimal has not been changed)


vm/float.go
```diff
@@ -444,7 +445,11 @@ func (f *FloatObject) lessThan(arg Object) bool {
 // ToString returns the object's value as the string format, in non
 // exponential format (straight number, without exponent `E<exp>`).
 func (f *FloatObject) ToString() string {
-	return strconv.FormatFloat(f.value, 'f', -1, 64) // fmt.Sprintf("%f", f.value)
+	s := strconv.FormatFloat(f.value, 'f', -1, 64)
+	if strings.IndexAny(s, ".") == -1 {
```
st0012: Can you add a comment to explain this case?

hachi8833: Added a comment, and revised the line to be more clear.

PR #731, opened 2018-08-31
Unify error messages in class.go and file.go

hachi8833:
Added tests for File class as well. 
Finally I will squash the commits before merging.

### DIFF
--- a/goby_test.go
+++ b/goby_test.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"github.com/goby-lang/goby/vm"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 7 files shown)


goby_test.go
```diff
@@ -2,17 +2,27 @@ package main
...
 	"testing"
+
+	"github.com/goby-lang/goby/vm"
 )
 
-func init() {
+func TestMain(m *testing.M) {
+	setup()
+	ret := m.Run()
+	//if ret == 0 {
```
st0012: do we still need this?

hachi8833:
Thank you for approving this! Squashing and merging...

PR #734, opened 2018-09-06
Add API docs to class.go

hachi8833:


vm/class.go
```diff
@@ -941,6 +984,39 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
...
+		// while the code is executing, which allows the code access to the receiver's
+		// instance variables and private methods.
+		//
+		// No other arguments can be taken.
+		//
+		// ```ruby
+		// string = "String"
+		// string.instance_eval do
+		//   def new_method
+		//		 self.reverse
```
st0012: can you fix the indentation here?

vm/class.go
```diff
@@ -941,6 +984,39 @@ func builtinClassCommonInstanceMethods() []*BuiltinMethodObject {
...
+		//   end
+		// end
+		// string.new_method  #=> "gnirtS"
```
st0012: is this on purpose?

hachi8833: Just a misindentation. I'll fix them.

vm/class.go
hachi8833: Just a misindentation. I'll fix them

PR #736, opened 2018-09-07
Fix error message in integer.go

hachi8833:
- Remove arg check for `to_d`

### DIFF
--- a/vm/integer.go
+++ b/vm/integer.go
@@ -396,9 +396,6 @@ func builtinIntegerInstanceMethods() []*BuiltinMethodObject {
 		{
 			Name: "to_d",
 			Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
-				if len(args) > 0 {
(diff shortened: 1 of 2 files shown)


vm/integer.go
```diff
@@ -396,9 +396,6 @@ func builtinIntegerInstanceMethods() []*BuiltinMethodObject {
 		{
 			Name: "to_d",
 			Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
-				if len(args) > 0 {
```
st0012: @hachi8833 why did you remove this?

hachi8833: Perhaps other `Integer#to_*` methods have no length checking and I thought this can be skipped. Should I revert this?

st0012: I tried on Ruby `2.4.3` and it raised an argument error:

```
irb(main):005:0> 100.to_d(100)
ArgumentError: wrong number of arguments (given 1, expected 0)
        from (irb):5
```

hachi8833: Thus I should add the checking args-length to other `Integer#to_*`. I'd update this PR.

PR #737, opened 2018-09-07
Unify error messages hash

hachi8833:


vm/hash.go
```diff
@@ -599,59 +616,63 @@ func builtinHashInstanceMethods() []*BuiltinMethodObject {
...
 				return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, "The value was not found, and no block has been provided")
-				
+
 			},
 		},
 		{
-			// Returns an array containing the values associated with the given keys but also raises
-			// ArgumentError when one of keys can’t be found.
+			// Returns an array containing the values associated with the given keys.
+			// When even one of keys can’t be found, it raises an ArgumentError.
```
st0012: Is the grammar of `When even one of keys can’t be found` correct?

PR #746, opened 2018-09-24
Add Array#index_with

hachi8833:
Ref: https://github.com/rails/rails/pull/32523


st0012:
@hachi8833 can you help me fix the failure?

PR #759, opened 2019-01-24
Fix #378: handle multiple double single quotation on igb REPL

### LINKED ISSUE #378: Should Use Newline when String Breaks Into Multiple Lines Else Returns Error
Goby REPL Log:

```ruby
$ go run goby.go -i 
# command-line-arguments
ld: warning: PIE disabled. Absolute addressing (perhaps -mdynamic-no-pic) not allowed in code signed PIE, but used in type..eqfunc.[106]string from /var/folders/70/jfl6pkgn39l1xxbg70r4cqdc0000gn/T/go-link-287248889/go.o. To fix this warning, don't compile with -mdynamic-no-pic or link with -Wl,-no_pie
Goby 0.1.1 😋 👹 😹
» puts "123"
123
#» nil
» "123
#» 12
» "1234
#» 123
» exit
Bye!
```

IRB Implementation:

```ruby
$ irb
2.3.3 :001 > "123
2.3.3 :002"> 123"
 => "123\n123" 
2.3.3 :003 > 
```

hachi8833:
Fixes #378.

Finally fixed the bug when multiple double/single quotations are in one line on igb REPL. Sorry for the delay🙇 @st0012 

   * expect: treats multiple line double quotation including single quotes

    ```ruby
    » if true
    ¤   if true
    ¤     puts "a", "a
    ¤ b's
    ¤ c"
    ¤   end
    » end
    a
    a
    b's
    c
    #»
    ```

* expect: treats multiple line double quotation including an open single quotes

    ```ruby
    » puts "a", "b'
    ¤ c
    ¤ d'
    ¤ e'
    » f"
    a
    b'
    c
    d'
    e'
    f
    #»

    » puts "a", "'b
    ¤ d
    ¤ b'
    » e"
    a
    'b
    d
    b'
    e
    #»
    ```

* expect: treats multiple line single quotation including double quotes

    ```ruby
    » if true
    ¤   if true
    ¤     puts 'a', 'a
    ¤ b "c"
    ¤ c'
    ¤   end
    » end
    a
    a
    b "c"
    c
    #»
    ```

* expect: treats multiple line single quotation including an open double quote

    ```ruby
    » puts 'a
    ¤ b"
    » c'
    a
    b"
    c
    #»

    » puts 'b', '"c
    ¤ d
    ¤ e"
    » d'
    b
    "c
    d
    e"
    d
    #»
    ```

### DIFF
--- a/igb/repl.go
+++ b/igb/repl.go
@@ -36,16 +36,16 @@ const (
 	reset     = "reset"
 
 	readyToExec = "readyToExec"
-	Waiting     = "waiting"
+	waiting     = "waiting"
 	waitEnded   = "waitEnded"
 	waitExited  = "waitExited"
 
-	NoMultiLineQuote     = "NoMultiLineQuote"
-	MultiLineDoubleQuote = "MultiLineDoubleQuote"
-	MultiLineSingleQuote = "MultiLineSingleQuote"
+	noMultiLineQuote     = "noMultiLineQuote"
+	multiLineDoubleQuote = "multiLineDoubleQuote"
+	multiLineSingleQuote = "multiLineSingleQuote"
 
-	NoNestedQuote = "NoNestedQuote"
-	NestedQuote   = "NestedQuote"
+	noNestedQuote = "noNestedQuote"
+	nestedQuote   = "nestedQuote"
 
 	emojis = "😀😁😂🤣😃😄😅😆😉😊😋😎😍😘😗😙😚🙂🤗🤔😐😑😶🙄😏😮😪😴😌😛😜😝🤤🙃🤑😲😭😳🤧😇🤠🤡🤥🤓😈👿👹👺💀👻👽🤖💩😺😸😹😻😼😽"
 )
@@ -147,24 +147,24 @@ reset:
 		sq := checkSingleQuoteOpen(igb.lines)
 
 		switch {
-		case igb.qsm.Is(NoMultiLineQuote):
+		case igb.qsm.Is(noMultiLineQuote):
 			switch {
 			case dq: // start multi-line double quote
-				igb.qsm.Event(MultiLineDoubleQuote)
+				igb.qsm.Event(multiLineDoubleQuote)
 				igb.startMultiLineQuote()
 				continue
 			case sq: // start multi-line single quote
-				igb.qsm.Event(MultiLineSingleQuote)
+				igb.qsm.Event(multiLineSingleQuote)
 				igb.startMultiLineQuote()
 				continue
 			}
 
-		case igb.qsm.Is(MultiLineDoubleQuote):
+		case igb.qsm.Is(multiLineDoubleQuote):
 			if igb.continueMultiLineQuote(dq) {
 				continue
 			}
 
-		case igb.qsm.Is(MultiLineSingleQuote):
+		case igb.qsm.Is(multiLineSingleQuote):
 			if igb.continueMultiLineQuote(sq) {
 				continue
 			}
@@ -199,8 +199,8 @@ reset:
 
 			// To handle beginning of a block
 			case pErr.IsEOF():
-				if !igb.sm.Is(Waiting) {
-					igb.sm.Event(Waiting)
+				if !igb.sm.Is(waiting) {
+					igb.sm.Event(waiting)
 				}
 				igb.printLineAndIndentRight()
 				continue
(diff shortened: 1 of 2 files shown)


igb/repl.go
```diff
@@ -377,6 +507,40 @@ func newIVM() (ivm iVM, err error) {
 	return ivm, nil
 }
 
+// Returns true if double quotation in the string is open.
+func checkDoubleQuoteOpen(s string) bool {
```
st0012: can't we extract duplicated logics from `checkDoubleQuoteOpen` and `checkSingleQuoteOpen`?

igb/repl.go
```diff
@@ -377,6 +507,40 @@ func newIVM() (ivm iVM, err error) {
...
+	s = strings.Replace(s, "\\'", "", -1)
+	sd, _ := openSingleQuote.MatchString(s)
+	if strings.Count(s, "'")%2 == 1 && sd {
+		return true
+	}
+	return false
+}
+
+// prompt switches prompt sign.
+func prompt(s int) string {
```
st0012: I think we should name this as an action, like `switchPrompt`

igb/repl.go
```diff
@@ -327,13 +352,98 @@ func handleParserError(e *parserErr.Error, igb *iGb) {
...
+	igb.cmds = append(igb.cmds, igb.lines)
+	println(prompt(igb.indents) + indent(igb.indents) + igb.lines)
+	igb.rl.SetPrompt(prompt2)
+}
+
 // Other helper functions --------------------------------------------------
 
+// filterInput just ignores Ctrl-z.
+func filterInput(r rune) (rune, bool) {
+	switch r {
```
st0012: we can just use `if` since there's only one case

igb/repl.go
```diff
@@ -327,13 +352,98 @@ func handleParserError(e *parserErr.Error, igb *iGb) {
...
+func (igb *iGb) continueMultiLineQuote(cdq bool) bool {
+	if cdq { // end multi-line double quote
+		igb.qsm.Event(NoMultiLineQuote)
+		if igb.nqsm.Is(NestedQuote) { // end nested-quote
+			igb.nqsm.Event(NoNestedQuote)
+			igb.cmds = append(igb.cmds, igb.lines)
+			println(prompt(igb.indents) + igb.lines)
+			igb.rl.SetPrompt(prompt2 + indent(igb.indents))
+			return true
+		} else { // exit multi-line double quote
```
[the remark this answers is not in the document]
st0012: @hachi8833 plz fix this

igb/repl.go
```diff
@@ -193,31 +224,28 @@ reset:
...
 				igb.caseBlock = true
 				continue
 
 			// To handle such as 'else' or 'elsif'
 			case pErr.IsUnexpectedToken():
-				igb.indents--
-				println(prompt2 + indent(igb.indents) + igb.lines)
-				igb.indents++
-				igb.sm.Event(Waiting)
+				println(prompt2 + indent(igb.indents-1) + igb.lines)
```
st0012: can you explain the calculation here in comment?

igb/repl.go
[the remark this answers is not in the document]
hachi8833: Fixed now

igb/repl.go
hachi8833: Extracted the two lines

igb/repl.go
hachi8833: You're correct. Updated.

igb/repl.go
hachi8833: I mean the prompt should be `¤` even on the top level indentation when the line is `else` or `elif`

```ruby
» if true
¤   99
¤ else
¤   88
» end
#» 99
```

```ruby
» if true
¤   99
¤ elsif true
¤   88
¤ elsif true
¤   77
¤ else
¤   66
» end
#» 99
```

st0012: Nice explanation, I hope you can put this in the comment though

igb/repl.go
st0012: I was hoping to see more extraction. The only differences of them are `^[^']*"`, `"\""` and `^[^"]*'`, `"'"`. I think we can make a generic function and pass these two differences as arguments to it.

hachi8833: Thank you for pointing this out! Integrated into one function.

igb/repl.go
hachi8833: updated

igb/repl.go
hachi8833: Commented

PR #760, opened 2019-02-02
Rename Goby to Gooby

jasonbarry:
It's Gooby! 

gooby


st0012:
Hmm, thanks but I don’t feel like changing it

PR #763, opened 2019-03-16
Integer#times should return an Enumerator

### LINKED ISSUE #762: `times` cannot yield an enumerator when chained to `map`
```ruby
# ruby
» 10.times.map do |i|
»   i * i
» end
#» [0, 1, 4, 9, 16, 25, 36, 49, 64, 81]
```

```ruby
# goby
» 10.times.map do |i|
¤   i * i
» end
#» InternalError: Can't yield without a block
```

st0012:
This is to fix #762 

The solution is kind of a workaround though: 
1) It's a `RangeEnumerator` instead of real Enumerator. 
2) The implementation is written in Goby instead of Go

So any suggestion is welcomed 😄


st0012:
@hachi8833 can you take a look at this?

hachi8833:
I will check this

Issue #751, opened 2018-10-31
Discussion: Goby's parameter syntax in method definitions


hachi8833:
TL;DR: I think we can simplify Goby's parameter syntax in method definition.

Of course the final decision should be @st0012's.

## Terms

Before proceeding, let me define the followings to clarify the issues and to be concise:

- **parameter**: on method definitions
- **argument**: on method calling

Name as the following on Goby's current method parameters:

- **pN**: normal parameters like `def foo(a)`
	- Unable to omit on method calling
	- No type restrictions
- **pND**: normal (optional) parameters with default value like `def foo(a=1)`
	- Can be omitted
	- no type restrictions
- **pK**: keyword parameters like `def foo(key:)`
  - Unable to omit on method calling
  - No type restrictions
- **pKD**: (optional) keyword parameters with default values like `def foo(key: "value")`
	- Can be omitted
	- No type restriction
- **pVA**: variable array parameters like `def foo(*array)`
	- Uses splat `*` for indicating pVA
	- Assumes an array will be taken
	- Can be defined only once in a method definition
	- Can be omitted (empty array `[]` will be passed when omitted)
	- No default value can be provided
	- "Bare" arguments will be treated as elements of the array

In Goby, the order of parameters are restricted as

1. pN: zero or more
2. pND: zero or more
3. pK: zero or more
4. pKD: zero or more
5. pVA: zero or one

Note that the order of arguments for pK can be shuffled as far as they are grouped:

```ruby
# Goby
def foo(key1:, key2:, key3:)
end

foo(key3: "foo", key2: "bar", key1: "baz")
```

This is the same for pKD:

```ruby
# Goby
def foo(key1: "foo", key2: 99, key3: [])
end

foo(key3: [55, 44, 33], key2: 88, key1: "bar")
```

### Ruby's parameters

In addition, Ruby has the following parameters:

- **pVH**: variable hash parameters like `def foo(**hash)`
	- Uses double splat `**` for indicating pVH
	- *Only hash can be taken*(!)
	- Can be defined only once in a method definition
	- Can be omitted (empty hash `{}` will be passed when omitted)
	- No default value can be provided
	- "Bare" keyword arguments will be treated as the elements of the hash
- **pB**: block parameters like `def foo(&block)`
	- Uses ampersand `&` for indicating pVH
	- Assumes `Proc` object will be taken
	- Can be defined only once in a method definition
	- No default value can be provided

### 1. Ruby's keyword parameters/arguments are still not well-integrated

[...]

PR #767, opened 2019-04-12
Remove pND params

hachi8833:
See https://github.com/goby-lang/goby/issues/751


st0012:
@hachi8833 I think it's still too soon to remove them without keyword parameters working properly, so closing it now. Sorry 😢

PR #771, opened 2019-06-09
Migrate to Go modules

hachi8833:
Sorry for my absence.
I checked the migration to Go modules from Godep and looks nice. 
Go 1.12 or later is required, and `export GO111MODULE=on` (until Go 1.13).

How about this?


hachi8833:
@st0012 Updated the .travis.yml. I think Go 1.11 or earlier should be out-of-scope.

st0012:
@hachi8833 I wonder why this was working even if we didn't update `.travis.yml` 🤔

hachi8833:
@st0012 I guess Go 1.11 has already the Go modules feature.

st0012:
@hachi8833 If the migrate is compatible with previous Go versions, I think we should still support them in `.travis.yml`

hachi8833:
That makes sense. I updated .travis.yml

st0012:
@hachi8833 thanks for all the work ❤️

PR #773, opened 2019-06-25
Replace fmt.Sprintf with strconv

st0012:
This is a PR for testing an optimization technic.

According to [this article](https://stephen.sh/posts/quick-go-performance-improvements), `fmt.Sprintf` is less performant than `strconv`.

```
🍎 strfmt → go test -bench=. -benchmem
goos: darwin
goarch: amd64
pkg: github.com/sjwhitworth/perfblog/strfmt
BenchmarkStrconv-8      30000000            39.5 ns/op        32 B/op          1 allocs/op
BenchmarkFmt-8          10000000           143 ns/op          72 B/op          3 allocs/op
```

So I changed some hotspots' `fmt.Sprintf` usages with more low level operations. (We actually use `fmt.Sprintf` a lot but most for inspection methods, but I think there's no need to change them.)

### DIFF
--- a/compiler/ast/expressions.go
+++ b/compiler/ast/expressions.go
@@ -2,7 +2,6 @@ package ast
 
 import (
 	"bytes"
-	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


st0012:
It doesn't make obvious difference, so closing it.

PR #774, opened 2019-07-28
Fix a typo

284km:
s/reciever/receiver/


st0012:
@284km thanks for the PR 😄

PR #775, opened 2019-07-28
Implements kind_of? to builtinClassCommonInstanceMethods

284km:
When I was writing test code for goby.
I noticed that ruby has kind_of? method. goby doesn't have it.
I wanna use kind_of? method. so I implement kind_of? method.

```
$ irb
irb(main):003:0> [].kind_of?(Object)
=> true
```

```
$ goby -i
» [].kind_of?(Object)
» NoMethodError: Undefined Method 'kind_of?' for []
```

In case of Ruby. is_a? method is alias to kind_of?
https://github.com/ruby/ruby/blob/b40589802b6fe6ba5cbb783a43f513cb2e597844/object.c#L4318

But I don't know to create alias in goby implementation.
So I implemented kind_of? method as a separate implementation.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -953,6 +953,51 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
 			return FALSE
 		},
 	},
+	{
+		// Returns true if Object class is equal to the input argument class
+		//
+		// ```ruby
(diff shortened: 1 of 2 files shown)


hachi8833:
Could you take a look at this? @st0012

PR #778, opened 2019-09-12
Temporarily remove race condition check for tests

st0012:
That's not Goby's top concern at the moment as it still lacks of many features. And the cause of race condition should be a fundamental design flaw. We should revisit this in the future when we have enough resources/knowledged for it. Try to fixing it case by case won't help at all.


st0012:
@hachi8833 can you take a look at this when have time?

PR #779, opened 2019-09-18
Add desc for using goby in spacevim

wsdjeg:


hachi8833:
Thank you for the PR!
Could you please take a look? @st0012

PR #782, opened 2019-09-25
Improve the output of `Object#inspect`

hachi8833:
I revised the output of `Object#inspect` method from `<Instance of: Class>` to `#<Foo:824635618784 >` or `#<Foo:824635618784 @bar={ k: [4, 5, "str"] } @foo=[1, 2, 3] >` (in the case of instances with instance variables).

```ruby
# Goby's new `#inspect` and `#to_s`
» class Foo; attr_accessor :foo, :bar;end
#» Foo
» a = Foo.new; a.foo = [1, 2, 3]; a.bar = {k: [4, 5, "str"]}
#» { k: [4, 5, "str"] }
» a.inspect
#» #<Foo:824635618784 @bar={ k: [4, 5, "str"] } @foo=[1, 2, 3] >
» a.to_s
#» #<Foo:824635618784 >
```

Note that the outputs of `Object#inspect` and `to_s` in Ruby are different as follows:

```ruby
# Ruby's #inspect and #to_s
» class Foo
»   attr_accessor :foo, :bar
» end
» a = Foo.new; a.foo = [1, 2, 3]; a.bar = {k: [4, 5, "str"]}
#» {:k=>[4, 5, "str"]}
» a.inspect
#» "#<Foo:0x00007fe103935bc0 @foo=[1, 2, 3], @bar={:k=>[4, 5, \"str\"]}>"
» a.to_s
#» "#<Foo:0x00007fe103935bc0>"
```

### DIFF
--- a/vm/error_test.go
+++ b/vm/error_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+
+	"github.com/dlclark/regexp2"
 )
 
 type errorTestCase struct {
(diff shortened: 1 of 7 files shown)


vm/validate.go
```diff
@@ -83,13 +87,15 @@ func verifyNullObject(t *testing.T, i int, obj Object) bool {
 
 func verifyStringObject(t *testing.T, i int, obj Object, expected string) bool {
 	t.Helper()
+	var fuzStr string
 	switch result := obj.(type) {
 	case *StringObject:
-		if result.value != expected {
+		re, _ := regexp2.Compile("(?<=#<[a-zA-Z0-9_]+:)[0-9]{12}(?=[ ]>)", 0)
+		fuzStr, _ = re.Replace(result.value, "XXXXXXXXXXXX", 0, -1)
```
st0012: So we need to use a fixed number of "X"s? It'd be hard to remember how may Xs we need to put. How about a flag-like value, like `[ANY_VALUE]`? And in the future, we can have several flags like `[INTEGERS]` to help the specs (this one is not necessary, just make an example).

vm/error_test.go
```diff
@@ -448,3 +450,21 @@ func checkErrorTraces(t *testing.T, index int, evaluated Object, expectedTraces
...
+	}
+
+	if fuzzifyMessage(err.message) != expectedErrMsg {
+		t.Fatalf("At test case %d: Expect error message to be:\n  %s. got: \n%s", index, expectedErrMsg, err.message)
+	}
+}
+
+func fuzzifyMessage(message string) string {
+	re, _ := regexp2.Compile("(?<=#<[a-zA-Z0-9_]+:)[0-9]{12}(?=[ ]>)", 0)
+	fuzMsg, _ := re.Replace(message, "XXXXXXXXXXXX", 0, -1)
```
st0012: Same suggestion as for `verifyStringObject`

vm/error_test.go
```diff
@@ -159,33 +161,33 @@ func TestStackTraces(t *testing.T) {
 
 func TestNoMethodError(t *testing.T) {
 	tests := []errorTestCase{
-		{`a`, "NoMethodError: Undefined Method 'a' for <Instance of: Object>", 1},
+		{`a`, "NoMethodError: Undefined Method 'a' for #<Object:XXXXXXXXXXXX >", 1},
```
st0012: Think we should also add a case for class methods (`Foo.bar`). Can you add it?

vm/validate.go
hachi8833: I got it! I feel "XXX..." looks a bit clumsy😅 and I'd revise that.

vm/error_test.go
hachi8833: I'll do as that.

vm/error_test.go
hachi8833: OK, I'm now adding more tests to the PR so I'll include the cases.

hachi8833:
Thank you for the approval! Merging...

PR #785, opened 2019-09-28
Fixed Variable Syntax Bug

### LINKED ISSUE #783: [变量名不合法]Variable name is illegal
Before you submit the issue, please provide us following information:

- Your operating system
- How did you install Goby
- Goby version
- The value of `$GOBY_ROOT`
- Error messages

If you have trouble building Goby on your machine, please check `$GOPATH` and `$GOROOT`

123a = 10 => result 10
123   = 10 => result 10
[]=10 => result 10

我人都懵了，我猜parser没做好检查

ChinaHDJ:
修正变量命名规则错误的问题
```ruby
{} = 1
[] = 1
123 = 1
@123 = 1

```

fixes #783

### DIFF
--- a/compiler/lexer/lexer.go
+++ b/compiler/lexer/lexer.go
@@ -193,49 +193,51 @@ func (l *Lexer) NextToken() token.Token {
 		tok.Type = token.EOF
 		tok.Line = l.line
 	default:
-		if isLetter(l.ch) {
+		if isLetter(l.ch) || isDigit(l.ch) {
 			if 'A' <= l.ch && l.ch <= 'Z' {
 				tok.Literal = string(l.readConstant())
 				tok.Type = token.Constant
 				tok.Line = l.line
 				l.FSM.Event("initial")
 			} else {
-				tok.Literal = string(l.readIdentifier())
-				if l.FSM.Is("method") {
-					if tok.Literal == "self" {
-						tok.Type = token.LookupIdent(tok.Literal)
-					} else {
-						tok.Type = token.Ident
-					}
-					l.FSM.Event("initial")
+				var isIdentifier, result = l.readIdentifierOrNumber();
 
-				} else if l.FSM.Is("initial") {
-					tok.Type = token.LookupIdent(tok.Literal)
-					if tok.Literal == "def" {
-						l.FSM.Event("method")
-					} else {
+				if isIdentifier {
+					tok.Literal = string(result)
+					if l.FSM.Is("method") {
+						if tok.Literal == "self" {
+							tok.Type = token.LookupIdent(tok.Literal)
+						} else {
+							tok.Type = token.Ident
+						}
 						l.FSM.Event("initial")
+
+					} else if l.FSM.Is("initial") {
+						tok.Type = token.LookupIdent(tok.Literal)
+						if tok.Literal == "def" {
+							l.FSM.Event("method")
+						} else {
+							l.FSM.Event("initial")
+						}
 					}
+					tok.Line = l.line
+				}else {
+					tok.Literal = string(result)
+
+					tok.Type = token.Int
+					tok.Line = l.line
+					return tok
 				}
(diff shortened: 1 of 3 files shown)


ChinaHDJ:
@st0012

ChinaHDJ:
@hachi8833 @st0012 review一下

ChinaHDJ:
> arr = [1, 2, 3]
>         		@a, @b, c = arr
>         		@a

i have a question. i write this example code to the path project_root/lib/array.gb.

it can passed. and i in igb can get variable value.

hachi8833:
Sorry, my reproduction was incorrect😅.

Well, could you change the your name `master` of ` ChinaHDJ:master` to other one like `fix-#783` or like that first? Pulling your branch to my local PC merges to my local `master`.

ChinaHDJ:
> Sorry, my reproduction was incorrect😅.
> 
> Well, could you change the your name `master` of ` ChinaHDJ:master` to other one like `fix-#783` or like that first? Pulling your branch to my local PC merges to my local `master`.

Sorry, I don't quite understand what you mean. I am a git newbie

st0012:
@ChinaHDJ thanks for the PR, really appreciate your work 😄  
Just some thoughts:
- Looks like the fix breaks some test cases, can you take a look?
- I expect there to be test cases that cover the bug you tried to fix (like the ones you reported in the issue). This is to make sure we won't cause any unexpected regression in future changes (like the failed test case caught some changes from your fix).
- Not sure if it's necessary to add `compiler/utils`. Can you explain why we need it?
- From the perspective of abstraction, we probably shouldn't have the methods in `compiler/utils` at the parser level. Checking individual characters is lexer's job. For parser, we should only deal with the tokens returned by the lexer. If you found that we really need to add those functions in parser to make things work, it's very likely that our lexer isn't doing a great job 😞 
- Although I'm not 100% sure, the cause should be only the `parser` component, at least for the `123 = []` case. 
- Finally, I recommend the following workflow for fixing this bug:
  1. Add test(s) in `parser/expression_parsing_test.go` to reproduce the bug (`123 = []`). Can use [this](https://github.com/goby-lang/goby/blob/master/compiler/parser/expression_parsing_test.go#L129) as a reference.
  2. Change parser to fix the test.
  3. Add another test case `123a = []`
  4. Make another change to fix the case
  5. If you found out the cause was in lexer instead of the parser, add a test example in lexer's test
  6. Fix the lexer's test case
  7. Check if the `123a = []` parser test case has been fixed. If not, fix it at the parser level (this should be a lot easier since the root cause in lexer has been fixed)

hachi8833:
> Sorry, I don't quite understand what you mean. I am a git newbie
@ChinaHDJ Changing the branch name is not critical and you can skip it for now.
Just focus on st0012's suggestions🙇.

ChinaHDJ:
> * it's very likely that our lexer isn't doing a great job 😞

问题主要是是因为，词法解析器和parser的问题。词法解析器比如说123ewq = 1的时候。123则会被判定为数字。ewq判定为变量。也就导致123ewq=1能通过，但实际上只是ewq赋值。然后是[]=xxx能够赋值成功是因为parser里没有加default当可用的case都没匹配上的时候，也就足以说明该赋值变量是存在问题的。所以我在switch default添加了主动报错

utils我是希望能够提取一些以后可以用得到的工具方法，让方法可以多用。只要import了就行

st0012:
Can you reply comments in English please? Thanks

ChinaHDJ:
> Can you reply comments in English please? Thanks

你不是台湾人吗？怎么会看不懂中文。

st0012:
This is an open source project, which means there will be people from different countries seeing this PR and try to understand what we’re doing. For example, @hachi8833 is also a core maintainer of this project, and he’s a Japanese so right now he can’t understand what you’re saying. I want to make the project’s progress and contributions transparent so we can get as many people involved as possible.

PR #786, opened 2019-09-29
Return error when found invalid assignment value (currently ignore)

st0012:
This fixes certain invalid syntaxes reported in #783 like

```
123 = []
[] = []
{} = []
```

But cases like `123foo = []` are caused by lexer's whitespace-insensitive. Which means that it doesn't consider whitespace while reading characters. So in this case, it'll tokenize the code into `Int, Ident, Eq, LBracket and RBracket` instead of `Illegal, Eq, LBracket and RBracket`.

To make the lexer whitespace-sensetive, we need to do some refactoring or even redesign for our lexer. So I choose to pick the low hanging fruits first and address that in other PR.

### DIFF
--- a/compiler/ast/expressions.go
+++ b/compiler/ast/expressions.go
@@ -146,9 +146,9 @@ func (he *HashExpression) String() string {
 		pairs = append(pairs, fmt.Sprintf("%s: %s", key, value.String()))
 	}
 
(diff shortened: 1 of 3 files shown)


st0012:
@saveriomiroddi thanks for the quick review!

PR #787, opened 2019-09-30
support parse single if

ChinaHDJ:
support the following

a = "ChinaHDJ" if true => result "ChinaHDJ"

b = "ChinaHDJ" if false => result nil

this result is the same as ruby irb

### DIFF
--- a/compiler/parser/expression_parsing.go
+++ b/compiler/parser/expression_parsing.go
@@ -166,6 +166,10 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 
 	leftExp := parseFn()
 
+	if p.peekTokenIs(token.If) && p.curToken.Line == p.peekToken.Line {
+		return p.parseSingleLineIfExpression(leftExp)
+	}
+
 	/*
 		Precedence example:
 
--- a/compiler/parser/flow_control_parsing.go
+++ b/compiler/parser/flow_control_parsing.go
@@ -88,6 +88,30 @@ func (p *Parser) parseCaseCondition(base ast.Expression) *ast.InfixExpression {
 	return infix
 }
 
+func (p *Parser) parseSingleLineIfExpression(content ast.Expression) ast.Expression {
+	p.nextToken()
+	ifExpression := &ast.IfExpression{BaseNode: &ast.BaseNode{Token: p.curToken}}
+
+	ce := &ast.ConditionalExpression{BaseNode: &ast.BaseNode{Token: p.curToken}}
+	p.nextToken()
+
+	// exmple puts 1 if true
+	//create expression stat for the code [ puts 1 ] if true
+	stmt := &ast.ExpressionStatement{BaseNode: &ast.BaseNode{Token: p.curToken}}
+	stmt.Expression = content
+
+	//parse Condition
+	ce.Condition = p.parseExpression(precedence.Normal)
+	ce.Consequence = &ast.BlockStatement{BaseNode: &ast.BaseNode{Token: p.curToken}}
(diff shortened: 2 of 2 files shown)


ChinaHDJ:
@hachi8833 review

hachi8833:
Thank you for the PR!

@ChinaHDJ Actually, once I consult with st0012 if we could add the `if` modifiers (or inline conditions). He thinks that the `if` modifier is not preferable for Goby's concepts and normal `if` block is preferable. 

We should write more documents to clarify the concepts and directions. By now, I hope the following slide might help you to understand the concepts:

https://docs.google.com/presentation/d/1hTHVV79UfZTB9RgXDOvs9S-KtEtZhAkFVsQhgb3qKFs/edit#slide=id.g36c296baf0_2_6

ChinaHDJ:
> We should write more documents to clarify the concepts and directions. By now, I hope the following slide might help you to understand the concepts:

i want goby can Can be the same as ruby

ChinaHDJ:
> if we could add the `if` modifiers (or inline conditions).

In my daily ruby development I use it often single line if.
because my code just one line, i dont need a block.

st0012:
@ChinaHDJ thanks for the PR. Will take a look later.
One advice: just saying “@someone review” is not very polite. I suggest “asking” the person if he can review your code, like “@someone can you review my PR when you have time?” or so. People will be more willing to give you help. 
This suggestion applies to projects you work on in the future, not just Goby 🙂

ChinaHDJ:
> Thank you for the PR!
> 
> @ChinaHDJ Actually, once I consult with st0012 if we could add the `if` modifiers (or inline conditions). He thinks that the `if` modifier is not preferable for Goby's concepts and normal `if` block is preferable.
> 
> We should write more documents to clarify the concepts and directions. By now, I hope the following slide might help you to understand the concepts:
> 
> https://docs.google.com/presentation/d/1hTHVV79UfZTB9RgXDOvs9S-KtEtZhAkFVsQhgb3qKFs/edit#slide=id.g36c296baf0_2_6



> @ChinaHDJ thanks for the PR. Will take a look later.
> One advice: just saying “@Someone review” is not very polite. I suggest “asking” the person if he can review your code, like “@Someone can you review my PR when you have time?” or so. People will be more willing to give you help.
> This suggestion applies to projects you work on in the future, not just Goby 🙂

I am sorry, my English is not very good. So it's hard for me to express my

st0012:
@ChinaHDJ that’s ok, just a friendly reminder 🙂

hachi8833:
I'm not so good at English, too😆, and unable to use Traditional Chinese. 
Don't worry so much!

st0012:
@ChinaHDJ Thanks for the suggestion and PR!! Implement your own feature request is a very good practice in open source software 😄 

However, we (ok, maybe just me) already decided *not* to support inline condition to make syntax rule simpler. This has 2 benefits:
1. Simpler grammar rule means simpler parser. Because Goby uses hand-written parser, it's better to keep it as simple as possible. This point might change when we have more people to work on parser but at the moment we don't 😢 
2.  I personally hate making/discussing coding style. So as the designer of Goby, I want to support only one type of syntax if possible. For if statement, I think `if .... end` block is enough for writing readable programs. Golang and several other languages don't support inline condition and developers can still write go programs. And in order to get a consistent/simple coding style, I think the slight inconvenience is an acceptable cost.

Due to the above reasons, I will close your PR for now, sorry. The reasons are all *my personal preferences*, so you did nothing wrong on this. You did a really good job 👍. I hope this won't discourage you from contributing open source projects.

PR #792, opened 2019-10-15
Add apidocs to #tap and #dup

hachi8833:
Added the api documents to `#tap` and `#dup`.

Ref: #788 and #791 

I hope my understanding is correct, especially on `Hash#dup`.

### DIFF
--- a/vm/array.go
+++ b/vm/array.go
@@ -618,6 +618,35 @@ var builtinArrayInstanceMethods = []*BuiltinMethodObject{
 		},
 	},
 	{
+		// Performs a 'shallow' copy of the array and returns it.
+		// Any arguments are ignored.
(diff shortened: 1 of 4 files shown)


st0012:
@hachi8833 sorry that I should warn you about `object_id` doesn't actually function like Ruby. Even if you call it on the same object it returns different numbers every time. 
I will try to fix this issue first. It that works out well, we can merge this PR after the fix. If that didn't work, you'll need to use different examples in your doc 😞

hachi8833:
Thank you for the explanation! I'll try it after the fix.

st0012:
@hachi8833 the issue of `object_id` has been fixed and merged, feel free to reopen this PR!

PR #793, opened 2019-10-19
[Refactor] Use Token Factory Method

Maxwell-Alexius:
While inspecting the code in lexer, I noticed that the `newToken` method in lexer file violates the DRY rule. Thus, I come up with a way to change the token creation logic back to the token declaration file.

I'm not sure if there is a better approach, but overall, I hope this PR can help:
- Improve readability, follow DRY principle
- Token creation logic should be placed to where tokens are declared


st0012:
@Maxwell-Alexius nice work! thanks 😄

PR #795, opened 2019-10-26
Use equalTo() as a interface for equality comparison

st0012:
Instead of defining `==` and `!=` methods separately, we can use the same method definition and rely on the underneath go function to perform the comparison check.

This PR is a prerequisite for fixing our `object_id` issue.


st0012:
@saveriomiroddi thanks for the opinion! I don't plan to implement `equal?` method, at least not with that name. That's a rarely used method and its name is very misleading IMO. We might have a similar method with names like `same_as?` in the future though.

PR #796, opened 2019-10-27
Fix object_id method

st0012:
In the past, even if you call `object_id` on the exact same object, it returns different values every time. This PR is to fix that issue + some refactoring around object initialization functions.


st0012:
@saveriomiroddi ah you're right, my bad 😕

st0012:
@saveriomiroddi updated, will merge after build finished

PR #797, opened 2019-10-28
Add API docs for #tap and #dup

hachi8833:
Retry of #792.

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -868,6 +868,33 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
 		},
 	},
 	{
+		// Performs a 'shallow' copy of the receiver object and returns it.
+		// Any arguments are just ignored.
+		// The object_id of the returned object is different from the one of the receiver.
+		// Note that `#tap` just returns self without copying if the receiver is an immutable object
+		// such as Integer, Decimal, Float or Regexp.
+		// Note that the internal statuses(instance variables) of the objects
+		// are also copied.
+		//
+		// See also `Array#dup`, `String#dup`, `Hash#dup`.
+		//
+		// ```ruby
+		// a = "string"
+		// a.object_id  #» 824637261824
+		// b = a.dup
+		// b.object_id  #» 824637263168
+		//
+		// class Foo
+		//   attr_accessor :foo
+		// end
+		// a = Foo.new     #» #<Foo:824634338592 >
+		// a.foo = 3.14
+		// a.inspect       #» #<Foo:824634338592 @foo=3.14 >
+		// b = a.dup
+		// b.inspect       #» #<Foo:824635635168 @foo=3.14 >
+		// ```
+		//
+		// @return [Object] Same type as the receiver
 		Name: "dup",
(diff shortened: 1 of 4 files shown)


vm/class.go
```diff
@@ -870,6 +870,33 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
 		},
 	},
 	{
+		// Performs a 'shallow' copy of the receiver object and returns it.
+		// Any arguments are just ignored.
+		// The object_id of the returned object is different from the one of the receiver.
+		// Note that `#tap` just returns self without copying if the receiver is an immutable object
```
st0012: Do we need to explain `tap` here?

vm/class.go
```diff
@@ -1545,6 +1572,38 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
 
 		},
 	},
+	// Just evaluates a given block with the receiver and returns the receiver.
+	// #tap method literally "taps into" the method chain and
+	// good for inspecting method chains.
+	// Any arguments to the method are just ignored.
+	//
+	// ```ruby
+	// (1..10).tap do |x|
```
st0012: can also put

```ruby
a = (1..10)
a.tap do |x|
# omit the rest
puts(a) #=> (1..10)
```

to show the original object is not touched

vm/hash.go
```diff
@@ -426,6 +426,45 @@ var builtinHashInstanceMethods = []*BuiltinMethodObject{
 		},
 	},
 	{
+		// Performs a 'shallow' copy of the hash and returns it.
+		// Any arguments are ignored.
+		// The object_id of the returned object is different from the one of the receiver.
+
+		// Caveat: any keys of hash ARE also copied with different object ids for now.
```
st0012: This shouldn't be the case, @hachi8833 can you open an issue for this and I'll fix it?

vm/class.go
hachi8833: Redundant as pointed out 😓.

vm/class.go
hachi8833: Got it

vm/hash.go
hachi8833: OK!

st0012: This won't be solved in the near future. Can you update other comments and I'll merge it?

hachi8833: @st0012 how about this?

> 		// Performs a copy of the hash, including the keys and values, and returns it.
>		// Any arguments are ignored.
>		// The object_id of the returned object is different from the one of the receiver.
>		// Caveat: any keys of hash ARE also copied with different object ids for now.
>		// This comes from the fact that the string objects are NOT frozen in current Goby.
>		// TODO: Perhaps the `Hash#dup` will not copy keys when we change the implementation of string objects as frozen.

st0012: Think the TODO part is unnecessary here. Otherwise looks good.

hachi8833: Thnx!

st0012:
@hachi8833 thanks!

PR #798, opened 2019-10-28
Split lexer tests into smaller pieces

st0012:
GitHub's diff works surprisingly well for this PR 😂


hachi8833:
Good works! This makes the test code far easier to maintain❤️

PR #802, opened 2019-11-16
Tadpole operator to `String#match?`

hachi8833:
This is just my idea that the tadpole operator `=~` looks cryptic and is a legacy from Perl, so changing this to `match?` is better for clarification. 

Rubocop style guides also suggests that `=~` is not (partially) recommended: https://github.com/rubocop-hq/ruby-style-guide#no-perl-regexp-last-matchers

If unnecessary, just discard the PR.
Thank you,


hachi8833:
No problem! I leave this to you.

st0012:
@hachi8833 I think this can go out in version `0.1.13`. Will merge when build runs green.

hachi8833:
Thank you!

PR #803, opened 2019-11-16
Cleanup goby binary executable for the case that test fails

hachi8833:
Small fix. `make test` performs `go clean .` to cleanup the local goby binary executable when test fails.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ install:
 .PHONY: test
 test:
 	go test $(TEST_OPTIONS) ./...
-	go clean .
+	make clean
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ install:
 .PHONY: test
 test:
 	go test $(TEST_OPTIONS) ./...
+	go clean .
 
 .PHONY: clean
 clean:


st0012:
@hachi8833 can you provide some numbers so we can have a clearer view of the benefits from the improvement?

hachi8833:
The remained binary executable sometimes disturbs rerunning tests. Just to `rm -rf goby` resolves the issue, but I think this can be automated.

Makefile
```diff
@@ -18,6 +18,7 @@ install:
 .PHONY: test
 test:
 	go test $(TEST_OPTIONS) ./...
+	go clean .
```
st0012: we already have a clean command in the below, I think we can just use it?

hachi8833: I thought as well. I'm a bit lazy to run `make clean` 😅

st0012: I mean we should use `make clean` here so we don't need to duplicate the logic for cleaning.

hachi8833: I understood. I'd close the PR.

st0012: No.... I mean we can put “make clean” in this line, so we are reusing the logic defined below. Can you make the change and reopen the PR?

hachi8833: Too hastened 😅I reopen and update it

PR #804, opened 2019-11-26
Fix Ripper.instruction

hachi8833:
It's been a long time since `Ripper.instruction` got unstable. 
I finally found that removing `AnchorLine()` fixes the issue.


st0012:
@hachi8833 LGTM, but can we also add some small test cases like

```ruby
def foo
  10
end
```

so it'll be easier to understand how it works?

I think there should be

1. single method definition
2. single class definition
3. single method call
4. single method call with a block (like `[1, 2, 3].each do |i|`)
5. whatever you think might help

hachi8833:
Added the smaller cases @st0012

PR #805, opened 2019-12-01
Fix misspells

hachi8833:
Just small fixes


vm/class.go
```diff
@@ -36,14 +36,14 @@ type RClass struct {
 var externalClasses = map[string][]ClassLoader{}
 var externalClassLock sync.Mutex
 
-// RegisterExternalClass will add the given class to the global registery of available classes
+// RegisterExternalClass will add the given class to the global registration of available classes
```
st0012: Think `registry` will be a better option?

st0012:
@hachi8833 there's no change in the PR after merging the master. I guess this is a subset of #806? If so, can you close it?

vm/class.go
hachi8833: Sorry, I should fix this as suggested.

PR #817, opened 2019-12-12
Fix by golint on compiler/bytecode

hachi8833:


compiler/bytecode/instruction.go
```diff
@@ -157,7 +157,7 @@ func (as *ArgSet) Names() []string {
 	return as.names
 }
 
-func (as *ArgSet) FindIndex(name string) int {
+func (as *ArgSet) findIndex(name string) int {
```
st0012: If changing this function name didn't break anything, maybe we can remove it?

hachi8833: Good catch!

hachi8833: Removed `findIndex()` on https://github.com/goby-lang/goby/pull/817

Issue #810, opened 2019-12-11
Random numbers


lara:
Your operating system: gobylang/goby:v0.1.10
How did you install Goby: Docker
Goby version: v0.1.10
The value of $GOBY_ROOT: /go/src/github.com/goby-lang/goby
Error messages: `UndefinedMethodError: Undefined Method 'rand' for <Instance of: Object>`

There doesn't seem to be a way to generate random numbers.

st0012:
@lara yeah you're right, we don't have such method atm. do you consider adding one 😄

lara:
@st0012 sure!

hachi8833:
I think using the secure `crypto/rand` is good. `math/rand` is less secure.

PR #819, opened 2020-01-15
Add random number generation by mapping rand() to math/rand

lara:
We're using `math/rand`, because when we're later implementing `srand()`, we need it to support seeding, which `crypto/rand` does not provide.

Fixes https://github.com/goby-lang/goby/issues/810

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -1354,7 +1354,7 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
 				}
 				return t.vm.InitIntegerObject(rand.Intn(args[0].Value().(int)))
 			case 2:
-				_, minOk := args[1].(*IntegerObject)
+				_, minOk := args[0].(*IntegerObject)
 
 				if !minOk {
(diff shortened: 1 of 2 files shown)


vm/class.go
```diff
@@ -1308,6 +1309,22 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
...
 	},
+	{
+		Name: "rand",
+		Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
+			aLen := len(args)
+			switch aLen {
+			case 0:
+				return t.vm.initFloatObject(rand.Float64())
+			case 1:
+				return t.vm.InitIntegerObject(rand.Intn(args[0].Value().(int)))
```
st0012: we probably should check arguments' types first. I think `.(int)` will make Go panicking when users pass something like a string.

lara:
@st0012 Sorry for the late turnaround, could you take a look again? I'd appreciate any tips for a simpler way to do safeguarding given 2 arguments at a time.

st0012:
@lara thanks for the work, everything looks perfect now :-)

Regarding the argument checks, I'm afraid there's no better way to do it atm. We did try adding some helpers to solve the issue, but that didn't work out too well. However, I will open a PR for spiking again in the following days.

PR #826, opened 2020-02-02
Fix by golint vm7

hachi8833:
Same as #822, #823, #824, #825

### DIFF
--- a/vm/validate.go
+++ b/vm/validate.go
@@ -6,7 +6,7 @@ import (
 	"github.com/dlclark/regexp2"
 )
 
-// VerifyExpected is a erification helpers for testing
+// VerifyExpected is a verification helpers for testing
 func VerifyExpected(t *testing.T, i int, evaluated Object, expected interface{}) {
 	t.Helper()
 	if isError(evaluated) {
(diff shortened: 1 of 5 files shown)


vm/validate.go
```diff
@@ -6,8 +6,7 @@ import (
 	"github.com/dlclark/regexp2"
 )
 
-// Verification helpers
-
+// VerifyExpected is a erification helpers for testing
```
st0012: Looks like you missed a v for verification?

vm/vm.go
```diff
@@ -278,6 +278,7 @@ func (vm *VM) initConstants() {
 	vm.objectClass.constants["STDIN"] = &Pointer{Target: vm.initFileObject(os.Stdin)}
 }
 
+// TopLevelClass returns a specified class or a top level class
```
st0012: It returns a specified top-level class (stored under the `Object` constant, like `String`, `Integer`..etc.)

vm/validate.go
hachi8833: fixed

vm/vm.go
hachi8833: Updated

PR #832, opened 2020-03-01
Support Array.new(args)

### LINKED ISSUE #831: Support for Array.new with an argument
We were trying to do `Array.new(n)` to get an array with n elements, but it's not supported.

st0012:
This PR supports multiple ways to initialize an Array:

```ruby
Array.new(3) #=> [nil, nil, nil]
Array.new(3, true) #=> [true, true, true]
Array.new(3) do |i|
  i * 2
end #=> [0, 2, 4]
```

Closes #831


st0012:
@hachi8833 thanks 👍

PR #833, opened 2020-03-07
Document Goby's error handling

st0012:


vm/vm.go
```diff
@@ -171,11 +171,23 @@ func (vm *VM) ExecInstructions(sets []*bytecode.InstructionSet, fn string) {
 	cf.self = vm.mainObj
 	vm.mainThread.callFrameStack.push(cf)
 
+	// this is the final destination of a Goby error
```
hachi8833: Suggestion:
> To Goby maintainers:
This is the final destination of Goby errors at VM level and we don't deal with the errors themselves at this point.
We only decide how the user program should react to them.

vm/vm.go
```diff
@@ -171,11 +171,23 @@ func (vm *VM) ExecInstructions(sets []*bytecode.InstructionSet, fn string) {
...
+	// at vm level, we don't deal with the error itself
+	// we only decide how the program should react to it
 	defer func() {
 		switch err := recover().(type) {
+		// if the error is still a Go panic until this phase,
```
hachi8833: > If the error is a true Go panic, Goby can't handle it properly and we should just `panic()`.

vm/vm.go
```diff
@@ -171,11 +171,23 @@ func (vm *VM) ExecInstructions(sets []*bytecode.InstructionSet, fn string) {
...
+		// it means Goby can't handle it properly and we should just let it crash
 		case error:
 			panic(err)
+
+		// if the error is something we raise intentionally, e.g. an argument error
```
hachi8833: > If the error is one of the Goby's errors such as argument error, we need to handle it gracefully depending on the mode of execution.

vm/vm.go
st0012:
```suggestion
		// if the error is still a Go panic until this phase,
```

vm/vm.go
st0012:
```suggestion
		// if the error is one of the Goby's errors, such as argument error, we need to handle it depending on the mode of execution.
```

vm/vm.go
st0012:
```suggestion
      // here is the final destination of Goby errors at the VM level, and we don't deal with them at this point.
      // we only decide how the user program should react to them.
```

st0012:
@hachi8833 Thanks for the feedback! FYI, you can use GH's new feature: review suggestion (see the 6th step of [this](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request)), so the PR author can just commit them right away like I just did 😆

PR #838, opened 2020-03-29
Add arg checking helper

st0012:
- Add `vm.checkArgTypes` for checking built-in methods' argument types
- Replace most of the manual type checkings with the new helper
- Some refactorings for reducing local variable usages

### DIFF
--- a/vm/class.go
+++ b/vm/class.go
@@ -186,7 +186,7 @@ var builtinModuleCommonClassMethods = []*BuiltinMethodObject{
 			module, ok := args[0].(*RClass)
 
 			if !ok {
-				return t.vm.InitErrorObject(errors.TypeError, sourceLine, errors.WrongArgumentTypeFormat, "a module", args[0].Class().Name)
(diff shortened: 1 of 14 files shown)


vm/class.go
```diff
@@ -1347,22 +1349,19 @@ var builtinClassCommonInstanceMethods = []*BuiltinMethodObject{
...
+
 				return t.vm.InitIntegerObject(rand.Intn(args[0].Value().(int)))
 			case 2:
-				_, minOk := args[0].(*IntegerObject)
 
-				if !minOk {
-					return t.vm.InitErrorObject(errors.TypeError, sourceLine, errors.WrongArgumentTypeFormat, classes.IntegerClass, args[0].Class().Name)
-				}
-				_, maxOk := args[1].(*IntegerObject)
+				err := t.vm.checkArgTypes(args, sourceLine, classes.IntegerClass, classes.IntegerClass)
```
st0012: @lara here's the helper to simplify argument checks I came up with. Let me know your thoughts about it.

lara: This is a pretty nice helper especially since it looks like it can check the types of multiple arguments at once. I also love that it's being standardized.

PR #840, opened 2020-04-15
Drop Plugin class completely

st0012:
This feature isn't actively supported by Go and is very unstable (still very cool though). And more importantly, it makes Goby harder to build on other platforms. So I'm dropping it now.


st0012:
@saveriomiroddi it turns out to be a feature that a user wants, so will keep it a bit longer. Sorry for the false call 😬

Issue #839, opened 2020-04-15
support define_method define_singleton_method


yannick:
as discussed, would you nice if `define_method` and `define_singleton_method` would be exposed for some fancy metaprogramming fun

--- phonehome.gb
Before you submit the issue, please provide us following information:

- Your operating system
macOs
- How did you install Goby
go get
- Goby version
0.1.12
- The value of `$GOBY_ROOT`
empty

PR #842, opened 2020-04-18
Support define_method and define_singleton_method

st0012:
This PR implements #839 but also contains other fixes & improvements that are required for the task:
1. Fixes block compilation's `argSet` being nil issue.
2. Fixes `def` statement working incorrectly inside `instance_eval` issue.
3. Fixes singleton method definition gets overridden issue.
4. Reformats Ripper test cases' expectations to improve readability.


st0012:
@saveriomiroddi ah it's always a pain to format goby code inside a go file 😞

Issue #841, opened 2020-04-16
Some feedbacks on documentation


st0012:
These are the feedback from @yawniek on slack

> 1. the vm.New vm.New and v.ExecInstructions methods would profit from some minimal documentation, especially what all the "paths" actually mean. some of them seemed semi obvious..
> 2. same for RegisterExternalClass ExternalClass and vm.Method  albeit those are relatively clear
> 3. how can i load a class into toplevel so i dont need a require in the goby code?
> 4. how can i load a method into the toplevel space / current vm context?
> 5. in general a more clean example of whats done in db/plugin on how to interfer with go structs and their method would be super helpful, especially when it comes to handling and converting numbers,maps and slices.
> 6. also explaining the details of the receiver vm.Object, sourceLine int, t *vm.Thread, args []vm.Object) vm.Object signature and how to efficiently work with it would be helpful

PR #845, opened 2020-04-19
Document and refactor vm package

st0012:
This PR is for addressing the 2 feedback I got from #841 

> 1. the vm.New vm.New and v.ExecInstructions methods would profit from some minimal documentation, especially what all the "paths" actually mean. some of them seemed semi obvious..
> 2. same for RegisterExternalClass ExternalClass and vm.Method albeit those are relatively clear

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -63,11 +63,12 @@ type VM struct {
 	fileDir string
 	// args are command line arguments
 	args []string
-	// projectRoot is goby root's absolute path, which is $GOROOT/src/github.com/goby-lang/goby
-	projectRoot string
 
-	// libPath indicates the Goby (.gb) libraries path. Defaults to `<projectRoot>/lib`, unless
-	// DefaultLibPath is specified.
+	// Goby vm uses libPath to find standard libraries written in Goby
+	// the fallback order is:
+	// 1. $GOBY_ROOT/lib
+	// 2. /usr/local/Cellar/goby/VERSION/lib - installed via homebrew
+	// 3. $GOPATH/src/github.com/goby-lang/goby/lib - development environment
 	libPath string
 
 	channelObjectMap *objectMap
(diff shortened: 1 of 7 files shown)


st0012:
@yannick can you give this a look?

vm/vm.go
```diff
@@ -150,6 +128,40 @@ func (vm *VM) newThread() (t Thread) {
...
+
+	gobyRoot := os.Getenv("GOBY_ROOT")
+
+	if len(gobyRoot) == 0 {
+		// fallback to homebrew's path
+		gobyRoot = fmt.Sprintf("/usr/local/Cellar/goby/%s", Version)
+
+		_, err := os.Stat(gobyRoot)
+
+		// if it's not installed via homebrew, find the
```
saveriomiroddi, line 146: "find the..."??? :grimacing:

vm/vm.go
```diff
@@ -150,6 +128,40 @@ func (vm *VM) newThread() (t Thread) {
...
+		return
+	}
+
+	gobyRoot := os.Getenv("GOBY_ROOT")
+
+	if len(gobyRoot) == 0 {
+		// fallback to homebrew's path
+		gobyRoot = fmt.Sprintf("/usr/local/Cellar/goby/%s", Version)
+
+		_, err := os.Stat(gobyRoot)
```
saveriomiroddi, line 144: I'm not sure what's the standard here, but I was wondering if the standard Go pattern is:

```golang
if _, err := os.Stat(gobyRoot); err != nil {
  // ...
}
```

vm/vm.go
```diff
@@ -150,6 +128,40 @@ func (vm *VM) newThread() (t Thread) {
...
+	if DefaultLibPath != "" {
+		vm.libPath = DefaultLibPath
+		return
+	}
+
+	gobyRoot := os.Getenv("GOBY_ROOT")
+
+	if len(gobyRoot) == 0 {
+		// fallback to homebrew's path
+		gobyRoot = fmt.Sprintf("/usr/local/Cellar/goby/%s", Version)
```
saveriomiroddi, line 142: I'd use `brew` commands to get the formula (I'm not familiar with Homebrew terminology) rather than using a hardcoded path, as it's a more generic (and robust) approach.

st0012: This is a good point. But I think that should be addressed in a different PR. Here's the issue for it https://github.com/goby-lang/goby/issues/846

PR #848, opened 2020-05-03
Make t.builtinMethodYield return an Object instead of a Pointer

st0012:
Most of the time, we use the return value of `t.builtinMethodYield` as an object instead of pushing it to the stack (as a Pointer). So it doesn't need to and shouldn't return a pointer, which costs us an extra attribute call in many cases.


st0012:
@saveriomiroddi thanks for the quick review!

saveriomiroddi:
> @saveriomiroddi thanks for the quick review!

This was a simple one :joy:

PR #849, opened 2020-05-03
Refactor ErrorObject and Error classes' initialization

st0012:
Currently, we allow raising errors from `vm` with any given string, like

```go
t.vm.InitErrorObject("ErrorWithoutAClass", ....)
```

 This is a bad practice because any internal error should have a corresponding error class.

This PR fixes this by only accepting pre-defined error types (enum) when initializing error objects. It also checks every error type (int) has a corresponding name (string) during vm initialization, which saves us the work to visually sync the 2 lists manually.


vm/error.go
```diff
@@ -22,7 +22,6 @@ type Error struct {
 	message      string
 	stackTraces  []string
 	storedTraces bool
-	Type         string
```
st0012, line 25: The `Type` attribute is essentially the error's class name. So we don't need to store it separately.

PR #862, opened 2020-10-22
Add URI.parse argument check

aisk:


vm/uri.go
```diff
@@ -21,6 +22,14 @@ var builtinURIClassMethods = []*BuiltinMethodObject{
 		// ```
 		Name: "parse",
 		Fn: func(receiver Object, sourceLine int, t *Thread, args []Object, blockFrame *normalCallFrame) Object {
+			if len(args) != 1 {
+				return t.vm.InitErrorObject(errors.ArgumentError, sourceLine, errors.WrongNumberOfArgument, 1, len(args))
+			}
+			typeErr := t.vm.checkArgTypes(args, sourceLine, classes.StringClass)
```
st0012, line 28: nitpick: can you add newlines before/after this line? we usually add newlines around the if statements for readability 😄

st0012 (commented):
thanks for the fix! just added a style related comment and I'll merge it once fixed 👍

vm/uri.go
aisk: Updated.

PR #864, opened 2020-10-25
Check is the codes is formated in travis ci

aisk:


aisk:
For go's format rules, I think if we continues using "Merge pull request" button to merge PRs, it will make the master's codes been un-formatted again. I think we should using "Rebase" method to submit PRs, and rebase every PR before submit to avoid this.

aisk:
I saw goby is using go module and the old `vendor` folder still here, and having code not been formatted, so just removed this folder, I don't know whether this is suitable?

st0012:
I understand the intention behind this change, but I don't feel the need to format all the code, especially not the vendor files. We use vendor because unlike normal applications, Goby's dependencies should be as steady as possible. And I want to minimize the risk that if a dependency is removed from the source, we won't be able to build our source code again.

PR #866, opened 2020-11-04
Add missing delete / options method to http module

aisk:


vm/http_test.go
```diff
@@ -37,7 +37,19 @@ func TestHTTPRequest(t *testing.T) {
 
 		res = Net::HTTP.head("http://127.0.0.1:3000/index")
 		res["Content-Length"]
-		`, "15"},
+		`, "16"},
```
st0012, line 40: can we use the exact content instead of the length of the content for comparison?

st0012 (commented):
great work, just a small comment 👍

vm/http_test.go
aisk: Currently head / delete / options only returns HTTP headers. I have post a thought on slack.

st0012: ah of course, my fault 🤦 
can we just test the status code then (should all be `204`)

aisk: Sorry late for the reply, just changed it to check the method from response headers.

st0012 (approved):

st0012:
@aisk sorry that I forgot to click merge after approving it 😬 mergint now

